PR #423, opened 2015-02-02
Fixed ons-tabbar to be hidden when set hideTabs

atakayama:
hideTabs flag didn't work when tabbar is located on the bottom of the screen

### DIFF
--- a/app/lib/onsen/js/onsenui.js
+++ b/app/lib/onsen/js/onsenui.js
@@ -12295,9 +12295,9 @@ limitations under the License.
           }
         } else {
           if (this._scope.hideTabs) {
-            this._tabbarElement.css('bottom', '0px');
+            this._contentElement.css('bottom', '0px');
           } else {
-            this._tabbarElement.css('bottom', '');
+            this._contentElement.css('bottom', '');
           }
         }
       },


argelius:
Hello @atakayama ,

Thanks for the pull request?

Before I merge this, could you please provide an example where `hideTabs` doesn't work? 

It seems to work fine in this example:
http://codepen.io/argelius/pen/xbXLjZ

atakayama:
Hello @argelius,

Sorry for lacking information.
As we can see in your example above, exactly tabbar becomes invisible by clicking "hide" button.
However, if any contents exist behind the tabbar, they won't appear.
It's like applying "visibility: hidden" on tabbar element.

In contrast, in the case if the tabbar is located on the top, we can use full screen when hiding the tabbar.
It works like "display: none".

argelius:
I see it now!

Thanks for the fix. I'll merge this now.

argelius:
I just realized the change was made in onsenui.js. That file is actually automatically generated, please make changes in the `framework` directory if you make another pull request.

I've moved the change to the correct place.

atakayama:
Sure, thank you

PR #432, opened 2015-02-06
Lazy repeat

argelius:
@anatoo さん

iOSだとまだ問題があります。iOSでスクロールしている間、全然レンダリングしてくれないです。
強制的にレンダリングをさせてみましたが、どんな方法でもスクロールが終わってからレンダリングします。

### DIFF
--- a/framework/views/lazyRepeat.js
+++ b/framework/views/lazyRepeat.js
@@ -104,6 +104,24 @@ limitations under the License.
       
       _renderElement: function(item) {
         if (this._isRendered(item.index)) {
+          // Update content even if it's already added to DOM
+          // to account for changes within the list.
+          var currentItem = this._renderedElements[item.index];
+
+          if (this._delegate.configureItemScope) {
+            this._delegate.configureItemScope(item.index, currentItem.scope);
+          }
+          else if (this._delegate.createItemContent) {
+            var newContent = angular.element(this._delegate.createItemContent(item.index)),
+              oldContent = currentItem.element.children();
+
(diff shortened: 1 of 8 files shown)


argelius:
@anatoo 

I changed so it will re-render the content every time it's updated so it can be used with `ons-pull-hook`. 
In the Chrome CPU profiler i get 80% idle time so it seems fast enough.

anatoo:
Great! 
I'm going to review your code and write Japanese doc comments.

argelius:
どうもありがとうございます

framework/views/lazyRepeat.js
```diff
@@ -0,0 +1,282 @@
...
+        $document[0].addEventListener('scroll', this._bindedOnScroll, true);
+      },
+
+      _removeEventListeners: function() {
+        $document[0].removeEventListener('scroll', this._bindedOnScroll, true);
+      },
+      
+      _destroy: function() {
+        this._removeEventListeners(); 
+        this._element = this._scope = this._attrs = null;
```
anatoo: Request: It is seem that "this._parentElement" and "this._renderedElements" is not released. Could you confirm that it is no DOM leaks?

Hirbod:
Seems this PR is just usable for list-items, not for normal content. That's a bit sad.
Maybe you could a small look on this:
https://github.com/lorenzofox3/lrInfiniteScroll

I've managed to get this work very well inside of OnsenUI

argelius:
It should work fine for any element as long as you know their height.

PR #435, opened 2015-02-09
Removed alias stack and added ons.componentBase variable.

argelius:


Hirbod:
What about deprecation-time? I've used ons.navigator 1-2 times

argelius:
@Hirbod We would not like to use the `ons.*` variables anymore since they're prone to some strange behavior. Also they've not been mentioned in the docs for several versions.

Hirbod:
This PR broke my complete app.. I know it is not so "nice", but sometimes, if I did changes on page 2, I want to do something on page 1 (e.g. on popPage)

```ons.page._scope.init();```

I don't know how to access this right now..

**Comment by Hirbod** [comment]
And you know what... it was such a big big big help when developing and testing inside of safari inspector. To get $scope vars, look for functions and make quick tests. My heart is bleeding you've removed it.. I really need a new to access these things

**Comment by Hirbod** [comment]
This is the real dirty way right now:

```
            angular.element($('ons-page:last')).scope().init();
```

argelius:
I'm really sorry about that @Hirbod,

Since it was an undocumented feature and it has some side effects we decided to get rid of it. I understand that some people we're using it but I think that's an even better reason to remove it since we don't want to keep supporting undocumented features.

I guess it would have been better to remove it when going from 1.1 to 1.2. Sorry about the trouble.

PR #491, opened 2015-03-11
remove an extra back quote

shingo-kumagai:
remove an extra back quote


argelius:
Thanks @shingo-kumagai !

PR #515, opened 2015-03-16
fastclick library update

andipavllo:


argelius:
Thanks @andipavllo !

Please do some tests to see if this works well with Windows Phone tomorrow.

argelius:
@andipavllo Please also confirm that it works well on Android and iOS

Issue #511, opened 2015-03-16
Can't control animation in TabbarView.setActiveTab(index, options) method


marcel-dancak:
'animation' attribute in the 'options' object has no effect, but by documentation it should be possible to control the animation.

PR #516, opened 2015-03-16
Fixed issue #511.

marcel-dancak:
Allow 'animation' option in TabbarView.setActiveTab(index, options) method.


argelius:
Thanks @marcel-dancak !

PR #526, opened 2015-03-20
Hammer default behaviour change + pullHook fix

andipavllo:
tested successfully in iOS, Android and Windows Phone 8.1

### DIFF
--- a/framework/js/onsen.js
+++ b/framework/js/onsen.js
@@ -234,6 +234,7 @@ window.ons = (function(){
   waitDeviceReady();
   waitOnsenUILoad();
   initAngularModule();
+  changeHammerDefault();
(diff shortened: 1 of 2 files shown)


framework/js/onsen.js
```diff
@@ -284,6 +285,10 @@ window.ons = (function(){
     });
   }
 
+  function changeHammerDefault() {
+    Hammer.defaults.behavior.touchAction = 'none';
```
argelius: Please add comment to explain why this is necessary.

argelius:
@andipavllo Looks good. Please add the comment and I will merge.

andipavllo:
@argelius comment added

PR #543, opened 2015-03-27
Add animation option to `popPage()`

seanchas116:
Current `popPage()` method has no animation option and performs the same animation as specified in `pushPage()`.

Add 'animation' option to `popPage()` method to specify different animation.

```javascript
myNavigator.pushPage("page.html", {
  animation: "slide" 
});
myNavigator.popPage({
  animation: "lift" 
});
```


argelius:
Merged. Good job!

PR #549, opened 2015-03-30
Hide empty notifications

zumbrunnen:
Notifications (badges) should be hidden if they have no content.

... and updated repo URL in README :wink:


argelius:
Thanks @zumbrunnen !

Didn't realize the repo URL was incorrect. That was embarrassing ;)

PR #550, opened 2015-03-30
Correctly removes the loading placeholder and works with templates. Fixes #541

### LINKED ISSUE #541: ons-loading-placeholder doesn't work for templates

frankdiox:


argelius:
@frankdiox, could you also write an e2e test for the loading placeholder?

PR #576, opened 2015-04-03
Created ons-col and ons-row elements as a custom element and removed ons-row and ons-col directives.

anatoo:
@argelius After v1.3.0 release, review and merge this.


argelius:
Looks great! Let's write all components as custom elements :)

PR #582, opened 2015-04-07
Custom animation with 'animation-options' attribute and 'animationOptions' parameter.

anatoo:
@argelius Please write missing English doc comments.


argelius:
@anatoo I've added the translations and updated the changelog.

anatoo:
@argelius Thanks!

PR #586, opened 2015-04-08
Lazy repeat fix

d3orn:
Minor fixed mainly for windows phone.
Sometimes this._parentElement[0] threw an error on windows when switching away from a LazyRepeat List or switching back to one.
This checks if this._parentElement is defined before getting the first entry.


argelius:
Thanks @d3orn ! I will take a look at this now.

argelius:
Looks great! Merged.

Thanks for all these fixes @d3orn ! :)

PR #598, opened 2015-04-15
Proposed spelling fixes

jsoref:
Hi, this pull request has a couple of segments:

* Branding fixes
* Generic spelling fixes
* Spelling corrections which may have some impact on consumers (these are identified w/ `[]`s)
* Miscellaneous (less likely to be correct)
* English US (potentially the wrong guess, if you happen to prefer a different English)

Please feel free to drop/squish things.
Each word-family is its own commit to enable fairly easy rebasing when repositories shift quickly, and to allow someone to drop any word-family which is intentional.


frankdiox:
@jsoref Thank you very much for all of these fixes! We will check them and merge :+1:

PR #606, opened 2015-04-17
if-platform now accepts multiple values

frankdiox:
Could you update the Japanese translation? :)


argelius:
@frankdiox Looks great! Please also update CHANGELOG.md

PR #609, opened 2015-04-20
Ons platform select

frankdiox:


core/lib/platform.es6
```diff
@@ -31,14 +45,22 @@ limitations under the License.
      * @return {Boolean}
      */
     isIOS: function() {
-      return /iPhone|iPad|iPod/i.test(navigator.userAgent);
+      if (!ons.platform._renderPlatform) {
```
argelius: Why do you check for the negative value? I think `if (ons.platform._renderPlatform)` is more clear.

argelius:
Thanks @frankdiox ! I've merged this now.

PR #654, opened 2015-05-14
ons platform detect whether device is Android mobile or tablet

JcDenton86:
@argelius is it ok?

### DIFF
--- a/core/lib/platform.es6
+++ b/core/lib/platform.es6
@@ -67,14 +67,14 @@ limitations under the License.
      * @return {Boolean}
      */
     isAndroidPhone: function() {
-        return /Mobile/i.test(navigator.userAgent);
+        return /Android/i.test(navigator.userAgent) && /Mobile/i.test(navigator.userAgent);
     },
 	
 	 /**
      * @return {Boolean}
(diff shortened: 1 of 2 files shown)


frankdiox:
@JcDenton86 thanks!
However, I think there is some mistake there since Tablet method is exactly the same than Phone method. Check it please :)

JcDenton86:
Hello @frankdiox. They are not the same as the Tablet function returns true if the userAgent has the String Android and not (!) the string Mobile. Check this fiddle: 

http://jsfiddle.net/wcjhzkhz/1/

Is that what you are talking about?

frankdiox:
@JcDenton86 Ohh sorry, I was looking for a more specific way to detect tablets (like reading 'tablet' in the user-agent) and did not see the '!' mark. Looks like this is the only way to detect tablets right now...
It will be reviewed probably on next week :)

frankdiox:
@JcDenton86 I have modified a couple of things and merged your commits in #660
Thanks!

JcDenton86:
Ok @frankdiox. Thanks!

PR #657, opened 2015-05-15
Added ons-toolbar element.

anatoo:


argelius:
Thanks @anatoo ! I couldn't find anything strange so I'm merging this.

PR #666, opened 2015-05-21
Added ons-switch element.

anatoo:


argelius:
Thanks @anatoo !

Looks good! I couldn't find anything and it works fine.

PR #693, opened 2015-05-28
ons-gesture-test

andipavllo:


argelius:
@andipavllo Is there some way you can extend this test to see if Hammer has been correctly initialized?

PR #698, opened 2015-05-29
Page life cycle events with specs

frankdiox:


frankdiox:
Finished!

argelius:
@frankdiox Please add this to the Changelog

argelius:
wow such commit much merge

anatoo:
Question: Page life cycle events can work with ons-split-view or ons-sliding-menu or other components except ons-navigator and ons-tabbar in this implementation?

frankdiox:
`init` and `destroy` event depends only on the page itself so they should work with any other element. `hide` and `show` are triggered only from Navigator and Tabbar. I think Sliding Menu does not "hide" the pages with `display: none`, it just takes the menu out of the visible screen, so it won't fire any event. Is there any other element that applies 'display: none' or 'display: block'?

anatoo:
"hide" and "show" event are not triggered on "ons-sliding-menu" or "ons-split-view" or using only "ons-page"? I think that this behavior confuse and annoy end-developers. I don't want to add this behavior on next release.

frankdiox:
Perhaps I haven't explained it well. It all happens in "ons-page-element", but only when CSS display is set to "none". Split-view and Sliding-menu don't set display to "none", they just take the page out of the screen. We can also fire "hide" and "show" events in those cases, but we just didn't speak about them in the specs.

anatoo:
I cannot prefigure this behavior.  When I read the spec, I think page lifecycle events must be triggered like the diagram below and CSS display property is not related always.
```
 create
   ↓
 show
  ↑ ↓
 hide
   ↓
 destroy
```

frankdiox:
`show` and `hide` events has been postponed: #725

PR #712, opened 2015-06-03
Fixed some bug.

zack24q:
Fixed error if event is undefined.
Fixed the wrong trigger when <ons-pull-hook> as a direct descendant of <ons-scroller>. issues#707

### DIFF
--- a/core/css/scroller.css
+++ b/core/css/scroller.css
@@ -17,6 +17,7 @@ limitations under the License.
 
 ons-scroller {
   display: block;
+  height: 100%;
   overflow: scroll;
   -webkit-overflow-scrolling: touch;   
   -ms-overflow-style: none;
@@ -24,4 +25,9 @@ ons-scroller {
 
 ons-scroller::-webkit-scrollbar {
   display: none;
+}
+
+.ons-scroller__content {
+  height: 100%;
+  overflow: scroll;
 }
--- a/framework/views/carousel.js
+++ b/framework/views/carousel.js
@@ -469,7 +469,7 @@ limitations under the License.
 
         event.stopPropagation();
 
-        this._lastDragEvent = event || null;
+        this._lastDragEvent = event;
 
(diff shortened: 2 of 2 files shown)


argelius:
Thanks @zack24q !

We will take a look at these fixes.

frankdiox:
Thanks @zack24q !
Have you encountered any bug related to undefined value in `event` variable? I believe in that case it should crash in line 470 before reaching your fix. Could you provide a example to reproduce the bug?

zack24q:
Thanks @frankdiox ,my "commented on e884a2b" was wrong, i look the code again and i found the error "Uncaught TypeError: Cannot read property 'gesture' of undefined" is throw by line 76. i look the code and i think this "commented 1f006ab" can fixed the error. Also i am writing an example  to reproduce this bug to you.

zack24q:
please see [codepen](http://codepen.io/zack24qq/pen/aOJjdy), i put ons-list in ons-carousel-item for requirement, sometimes it will throw that error when i scroll down the list.

frankdiox:
@zack24q I am sorry but after hundreds of tries in different browsers I haven't been able to get the error message in your Codepen :(
Can you give me more details about how to reproduce it? Thanks!

zack24q:
@frankdiox my system is window 8 and my browser is chrome 42, and i found my reproduce way.
my network is bad, and i open the Codepen, the result-iframe in Codepen will spend several seconds to load, when it is loaded, i scroll down the list immediately, the error will throw out, and after scrolling several times, there is no new error be thrown.
Another time, when it is loaded for a while, i scroll down the list and there is no new error be thrown.
so i think maybe something is not initialization completed, so user scroll the list will throw that error.
i hope this will help you.

frankdiox:
@zack24q Finally I got the error! I've made another pull request with your changes, but I modified them a little bit. We are also using a specifing commit message format now. Check it out here: #719
Thanks!

PR #713, opened 2015-06-03
fix(ons-popover): popover demo

frankdiox:


demo/popover/index.html
```diff
@@ -12,6 +12,8 @@
     <script src="../app.js"></script>
```
argelius: Was the var not created in this file?

demo/popover/index.html
```diff
@@ -12,6 +12,8 @@
     <script src="../app.js"></script>  
 
     <script>
+      var app = angular.module('myApp', ['onsen']);
```
argelius: I think you should do `angular.module('myApp')` since the app is already initialized.

frankdiox:
It creates the module but does not assign to any variable. Other demos do the same.

PR #726, opened 2015-06-11
perf(ons-page): Copy child nodes instead of innerHTML.

argelius:
@anatoo 

Copying innerHTML caused createdCallback to be triggered again for all children.

### DIFF
--- a/core/elements/ons-page.es6
+++ b/core/elements/ons-page.es6
@@ -149,10 +149,8 @@ limitations under the License.
       var content = document.createElement('div');
       content.classList.add('page__content');
 
-      var html = this.innerHTML;
-      content.innerHTML = html;
-      while (this.childNodes[0]) {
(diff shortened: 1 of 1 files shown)


anatoo:
@argelius I initially implemented with removeChild and appendChild instead of innerHTML. But it caused angular directive error on "demo/navigator" with iOS and Chrome. Does your implementation work on demo?

argelius:
I'll check it out!

anatoo:
If there is no problem, your implementation is better.

argelius:
It fails on iOS. I will investigate this.

I still think we need to do something different than copying innerHTML since createdCallback will be called far too times for nested elements.

argelius:
@anatoo This implementation works.

anatoo:
Good! Merged it.

PR #734, opened 2015-06-18
Fixed ReferenceError: MutationObserver and CustomEvent are not defined.

ta2xeo:
Android doesn't work in less than OS version 4.4.
MutationObserver.js copied from "https://github.com/webcomponents/webcomponentsjs/blob/master/src/MutationObserver/MutationObserver.js"
verification devices
* AQUOS PHONE ZETA SH-01F NTT docomo (OS 4.2.2)
* Galaxy SⅢα SC-03E NTT docomo (OS 4.3)

### DIFF
--- a/core/vendor/CustomElements.js
+++ b/core/vendor/CustomElements.js
@@ -629,7 +629,7 @@ CustomElements.addModule(function(scope) {
       }));
     });
   }
-  if (isIE11OrOlder && typeof window.CustomEvent !== "function") {
+  if (typeof window.CustomEvent !== "function") {
     window.CustomEvent = function(inType, params) {
       params = params || {};
       var e = document.createEvent("CustomEvent");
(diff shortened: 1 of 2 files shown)


argelius:
Hi @ta2xeo !

We've updated the polyfill to a newer version. Can you test the master branch again and see if it works?

ta2xeo:
Thank you for reply.

"CustomEvent" doesn't work.

console log below

```
E/Web Console﹕ Uncaught ReferenceError: CustomEvent is not defined:1109
E/Web Console﹕ Uncaught ReferenceError: CustomEvent is not defined:9358
E/Web Console﹕ Uncaught ReferenceError: CustomEvent is not defined:10349
```

I guess that CustomEvent is not implemented on several old android devices.

Should I define "CustomEvent" in my program?

argelius:
Hi @ta2xeo !

Sorry for the delay. I've added a CustomEvent polyfill now. Does it work for you now?

ta2xeo:
Thanks @argelius !

It works now!
I close this PR.

argelius:
Thanks @ta2xeo !

PR #739, opened 2015-06-24
Added ons-back-button custom elements.

anatoo:
Merge this changes after merging ons-navigator-element if there is no problem.


argelius:
Looks good!

PR #753, opened 2015-07-01
fix(ons-tabbar): fixed loadPage() bug. Close #751

### LINKED ISSUE #751: Ons-tabbar loadPage() function doesn't destroy previous page
The page stack increase every time the function has been called, it should not be default behavior.

andipavllo:


argelius:
Thanks @andipavllo !

PR #778, opened 2015-07-14
fix(onsenui.js): fixed ons.bootstrap() on Windows, fixes #777

### LINKED ISSUE #777: ons.bootstrap() gives exception in Windows
In windows app, ons.bootstrap() does not work.
I found that the doc.readyState is 'interactive' when error is occurred.
In such case, angular.bootstrap should be called after DOMContentLoaded event is fired
same as the case doc.readyState is 'loading' or 'uninitialized'.

Please modify
        if (doc.readyState == 'loading' || doc.readyState == 'uninitialized') { ...
 in ons.bootstrap method by
        if (doc.readyState == 'loading' || doc.readyState == 'uninitialized' || doc.readyState == 'interactive') { ...
.

andipavllo:


argelius:
@andipavllo Please commit the fix only.

PR #785, opened 2015-07-16
Moved some ons.*() methods to core.

anatoo:


argelius:
Thanks @anatoo! Will merge now.

PR #807, opened 2015-07-24
Hold in hammer 2 not work. now there is press.

carlbaubau:


argelius:
Hi @carlbaubau !

Currently, on the master branch we are using a modified version of Hammer. This is based on an older version that has `hold` instead of `press` so the `<ons-gesture-detector>` will not use work with `press`.

We have renamed the Hammer object in order to avoid conflicts if another version is loaded.

If you want to use Hammer 2 with Onsen UI 1.3.x you need to rename the variable. However, we will release a version with a renamed Hammer object shortly.

PR #809, opened 2015-07-24
Navigator code coverage and minor fix

frankdiox:
Fixed the behaviour of `insertPage()` method when the provided index is higher or lower than the number of pages in the stack.


argelius:
Thank you @frankdiox !

PR #816, opened 2015-07-28
Tabbar code coverage and Tabbar fix

frankdiox:


argelius:
Thank you @frankdiox !

PR #833, opened 2015-08-06
Page lifecycle events

frankdiox:
### Page Lifecycle Events

This PR allows ons-pages to throw lifecycle events even if they are appended to nested components (tabbar -> navigator -> page).

* Every nested component will be wrapped inside an `ons-page` automatically in order to have the same API and make it homogeneous. The only exception is when the component is the main element (`<body><ons-navigator> ... </ons_navigator></body>`).
* Components must implement `_hide()`, `_show()` and `_destroy()`.
* Every `ons-page` will throw `show` and `hide` when showing or hiding. This also happens after its creation, if immediately shown, and before its deletion, if it was displayed at that moment.
* Page lifecycle events will be propagated to the page's descendants so they are correspondingly shown, hidden or destroyed. For example, destroying an `ons-sliding-menu` that contains an `ons-navigator` as the main page will throw `hide` event only for the displayed page but `destroy` event for every page in navigator's page stack.
* The wrappers mentioned in the first point will **not** throw events. There is a special `_muted` attribute for this.
* `ons-sliding-menu` and `ons-split-view` cannot implement this API since they are not WebComponents yet. This has been amended in [L102](https://github.com/OnsenUI/OnsenUI/pull/833/files#diff-eaf02449e8c64b9d64988657b8528a59R102). The only minor drawback is that menu-pages will also throw events.
* `ons-fab` would also use this API to hide and show the button.


frankdiox:
@anatoo @argelius before addressing the AngularJS bindings, what do you think about this?

anatoo:
It looks good to me. Great work!

frankdiox:
@anatoo @argelius I added angular directives for the events. I'm not sure if this is the way you wanted it. 
An page lifecycle event can be seen from the own page or any parent component:

```
<ons-navigator ... ons-hide="doSomethingElse()">
  <ons-page ... ons-hide="doSomething()">
     ...
  </ons-page>
</ons-navigator>
```

Also like this `$scope.myNavigator.on('hide', handler);`.

anatoo:
This changes will merge in https://github.com/OnsenUI/OnsenUI/pull/896.

PR #840, opened 2015-08-11
Inline badges

frankdiox:


argelius:
Agree this looks better since we have a lot of them.

PR #864, opened 2015-08-23
refactor(all): Fix typos to enable strict mode in some view codes.

fruitymidnight:
I found some typos in framework/views codes.
If  you want to enable strict mode, you must write "'use strict';" instead of "'use strict;';"


anatoo:
@fruitymidnight Your changes is merged. Thanks!

Issue #879, opened 2015-09-08
ons-navigator as a main page of ons-slide-menu flickering


c58:
II have three templates: one with ons-slide-menu, another(page) with ons-page and last with ons-navigator with default page attribute set.
When i navigate with slide menu to a page without navigator all is fine. But when i navigate to a page with ons-navigator page is flickering (black page before actual page).
Problem occurred on desktop Chrome and on IOS 9 (Cordova), others not tested.
There is a proof.
![page-flickering.gif](https://cloud.githubusercontent.com/assets/1871220/9728522/eaafe4ae-5621-11e5-9394-536d30786a94.gif)

c58:
Page one:
```
<ons-navigator animation="slide" var="pageNav"
  page="partials/app/selector/list.html">
</ons-navigator>
```

Page two:
```
<ons-page var="pageTimetable" ng-controller="TimetableController as ctrl">
  <ons-toolbar class="st-nav" modifier="android">
  ...
```

Page list.html:
```
<ons-page class="st-selector" var="pageSelector" ng-controller="SelectorController as ctrl">
  <ons-toolbar class="st-nav">
  ...
```

argelius:
Thanks for reporting @c58 !

This looks like a bug. We will investigate it.

frankdiox:
I can confirm the flickering.
In sliding menu: http://codepen.io/frankdiox/pen/VvLNpK
In tabbar: http://codepen.io/frankdiox/pen/EVjzag

We will try to fix it!

c58:
Thanks! Just for now i've made a workaround. Flickering is still there, but not so ick :)
<ons-navigator style="background: #fff;">
</ons-navigator>

Hirbod:
same flickering when having tabbar with own navigator. as soon as there are nested ons-pages, it will flicker. This is why I still have never implemented a second navigator (I lose my tab-bar on pushPage, I would love to keep it, but played around more than 10 hours to fix the flickerin.) no chance..

c58:
I think my strategy of the fix may be applied on tabbar too. I'll try to fix it, maybe in the evening.

c58:
I've investigated, that the problem is with using a promises in `getPageHTMLAsync` and `getTemplateHTMLAsync`. They should be synced when template already in a templates cache. Then a ons-navigator's main page will be initialized synchronously (within the same run cycle that _appendMainPage called) and no flickering will be appeared.

Hirbod:
[...]

PR #899, opened 2015-09-11
fixed navigator flickering in sliding menu #879

c58:
Only fixes a bug with sliding menu. I don't use a tab bar, so may be i fix it a little bit later.


argelius:
Thanks @c58 ! I've solved this in a slightly different way. Could you please verify that the flickering is gone?

https://github.com/OnsenUI/OnsenUI/commit/6c0dca0c57cba93ef3b96acd01885b3892a1d019

c58:
No, flickering is still there with your fix.
I think your fix is fifty-fifty, especially on mobile devices. There is a risk that between 40ms main page of navigator will not initialized. But my conception works in any situation.

argelius:
Thanks @c58 ! I will open this again.

PR #901, opened 2015-09-14
fix(ons-page): removed page__content margin

frankdiox:


argelius:
Hi! I can not find the discussion about this bug anymore. Could you paste an example of the bug that this solves?

If there is no problem I will merge tomorrow.

frankdiox:
@argelius Change tabs here and add the -1px to page__content: http://codepen.io/frankdiox/pen/EVjzag

I don't know if that -1px was fixing something else, I didn't find anything so perhaps it's not necessary for Onsen 2.0.

anatoo:
I added this line(margin-top: -1px) for a display issue on iOS. But I don't remember the details.
It's OK to remove this line if there is no problem on iOS preview.

argelius:
I made some tests on iOS7 and iOS8 without any trouble. Will merge this.

PR #903, opened 2015-09-14
Update carousel.js

astanecki:
Description update according to current tree state


argelius:
Thanks a lot @astanecki !

This applies to some other docs as well so I will update those later as well.

PR #896, opened 2015-09-11
Page lifecycle events integration

anatoo:
@argelius Merge this change after "ons-splitter" branch.

PR #906, opened 2015-09-16
Changed the way to trigger custom events.

anatoo:
This changes allows to refer custom event properties like the below:

```
navigatorElement.addEventListener('poppage', function(event) {
  console.log(event.detail.foobar); // old way
  console.log(event.foobar); // new way
});
```

@argelius Please review and merge this pull request after https://github.com/OnsenUI/OnsenUI/pull/896.


argelius:
Thanks @anatoo! I have confirmed this so I will merge it.
I agree this is better and it will not break backwards compatibility.

PR #907, opened 2015-09-16
feat(ons-navigator): bringPageTop( page | index,  [options]). Closes #877

### LINKED ISSUE #877: Feature request: navigator.bringToTop( page );
It checks if the provided page already exists in the navigator's page stack and brings it to the top in such case. If it doesn't exist it could use `pushPage()` as fallback.

It can also be implemented as a new option for `pushPage()`.

@argelius @anatoo, thoughts?

frankdiox:
This method brings an existing page to the top with push animation. This page can be provided by `pageUrl` or a numeric index. This closes #877 .

I have made some assumptions here. Let me know if we should do it in a different way:
* If the page doesn't already exist it falls back to `pushPage()`.
* It does nothing if the page is already on top.
* The provided `options` object will overwrite the previous one (the one assigned to the `pageObject` when the page was pushed the first time). Options are the same than for `pushPage()`.

Even though the page is already attached I had to reattach it to avoid animation bugs. Since we throw `init` events every time a page is attached to DOM I have added an attribute to avoid it. 

We need Japanese translations in `framework/directives/navigator.js`.


framework/directives/navigator.js
```diff
@@ -257,6 +257,29 @@
...
+ *   [ja]アニメーション名を指定します。"slide", "simpleslide", "lift", "fade", "none"のいずれかを指定できます。[/ja]
+ * @param {String} [options.animationOptions]
+ *   [en]Specify the animation's duration, delay and timing. E.g.  <code>{duration: 0.2, delay: 0.4, timing: 'ease-in'}</code>[/en]
+ *   [ja]アニメーション時のduration, delay, timingを指定します。e.g. <code>{duration: 0.2, delay: 0.4, timing: 'ease-in'}</code> [/ja]
+ * @param {Function} [options.onTransitionEnd]
+ *   [en]Function that is called when the transition has ended.[/en]
+ *   [ja]pushPage()による画面遷移が終了した時に呼び出される関数オブジェクトを指定します。[/ja]
+ * @description
+ *   [en]Brings the given page to the top of the page-stack if already exists or pushes it into the stack if doesn't.[/en]
+ *   [ja][/ja]
```
argelius: @anatoo Could you please add Japanese translation here when you have time?

anatoo: @argelius Ok! I added transition comments.

argelius:
@frankdiox I couldn't find any problems so I will merge this.

PR #908, opened 2015-09-17
Mutable animation-options attribute.

anatoo:
This changes enable to change animation-options attribute dynamically on all custom elements.
@argelius Merge this pull-request if there is no problem.


argelius:
Thanks @anatoo! Looks great. I can't find any problems so I will merge it now.

anatoo:
Thanks for merge!

PR #916, opened 2015-09-20
chore(dist): modified release script

frankdiox:
Pushes the current version to https://github.com/OnsenUI/OnsenUI-dist
`./dist-release.sh`
or
`./dist-release.sh no-build` (doesn't run `gulp build` before)


argelius:
Thanks @frankdiox ! Let's use this for the next release. I think we also need to tag old versions in this repo. Can you do that?

Also, could you add some documentation for this to README.md?

frankdiox:
@argelius I uploaded all the old versions after 1.1.0. Before that OnsenUI didn't even use Gulp, so this script won't work.

Where do you want the documentation, in OnsenUI/README.md or OnsenUI-dist/README.md?

argelius:
@frankdiox Please put the documentation in this repo but also write a README.md in the other repo.

frankdiox:
@argelius We need to change now bower and npm to point to OnsenUI-dist. Can you do that?

argelius:
Thanks @frankdiox !

PR #920, opened 2015-09-24
Material toolbar

argelius:
@anatoo I have added styles for a Material Design toolbar.

Please merge if there is no problem.

I added the Roboto webfont for the css components page so it renders correctly there.


anatoo:
@argelius  Thanks! I can't find any problems so I will merge this.

argelius:
Thanks @anatoo!

PR #933, opened 2015-10-02
Material Alert Dialog

argelius:


argelius:
Implemented the design and fixed the `ons.notification.*` methods.

@anatoo
I'm not sure about the classnames:

Right now it's like this:
```
alert-dialog--material-title--one
```

But I think this is better:

```
alert-dialog-title--material--one
```

Or even this:

```
alert-dialog--material__title--one
```

anatoo:
@argelius 
hmm... "alert-dialog--material-title--one" class name looks strange.
Could you fix this naming as you say?

argelius:
@anatoo I changed the class names.

anatoo:
Merged.

PR #937, opened 2015-10-05
Lazy repeat reload

argelius:
@anatoo I've added a `reload()` method to the Delegate object.

I also fixed a bug with the layout of the wrapper element. 

Also, I made it easier to put elements before the list. Now you can do

```html
  <ons-page>
    <ons-toolbar>
      ...
    </ons-toolbar>
    <p>Before list</p>
    <ons-list>
      <ons-list-item ons-lazy-repeat="Delegate">
        ...
      </ons-list-item>
    </ons-list>
  </ons-page>
```

Without breaking the layout.


anatoo:
@argelius Great work! Thanks!
Merged.

PR #941, opened 2015-10-09
Animit improved

anatoo:
Added some performance optimization and refactoring on animit.js.

 * Changed to use setImmediate() and requestAnimationFrame() on forcing synchronous layout for rendering performance.
 * Renamed resetStyle() to restoreStyle() and added saveStyle() that have corresponding function.
 * Animit object no longer use "data-animit-orig-style" attribute for saving style data. 
 * Added some optimization on wait() method.

@argelius Please merge this if there is no problem.


argelius:
@anatoo Looks great! Merged.

PR #943, opened 2015-10-13
feat(css-components): Add "material" modifier to "page" class.

argelius:
@anatoo 

This will add Roboto/Noto font to `<ons-page>` elements by using "material" modifier:

I think this is necessary in order to override default Helvetica font.

```
<ons-page modifier="material">
...
</ons-page>
```


anatoo:
@argelius Nice work! I think it's necessary too.
Merged.

PR #944, opened 2015-10-13
Material back button

argelius:
@anatoo

This will change the style of the back button depending on if the parent toolbar has the "material" modifier or not.

Please merge if there is no trouble.


anatoo:
@alexeygolev Good! Merged.

PR #953, opened 2015-10-15
fix(ons-tabbar): closes #948

### LINKED ISSUE #948: Error when switching tabs
In this demo 
https://github.com/OnsenUI/OnsenUI/blob/master/demo/tabbar/index.html

The following error is thrown sometimes:
```
Uncaught TypeError: Cannot read property '$destroy' of undefined
```

It can be triggered by clicking the tabs furiously.

frankdiox:


argelius:
@fran: Does this not cause a memory leak?

frankdiox:
@argelius The memory leak was in master before this PR. This fixes another error but does not fix the leak.

argelius:
@frankdiox Ok. Are you working on fixing the memory leak?

If not, please create an issue so we know how to reproduce it.

frankdiox:
@argelius I just reported it. The memory leak is much lower after this fix but there is still something else creating it.

PR #957, opened 2015-10-16
Fix 950

argelius:
@anatoo 

This fixes a bug with status bar filling on iOS9. The ~ selector doesn't work as expected in the new UIWebView so I have replaced with + selectors.


anatoo:
@argelius Ok! Merged.

PR #992, opened 2015-10-29
feat(ons.platform): uses cordova-plugin-device if installed. Closes #842

### LINKED ISSUE #842: Use "cordova-plugin-device" in ons.platform methods.
Use the plugin when detecting device instead of user agent sniffing. Fallback to user agent if the plugin is not present.

frankdiox:


argelius:
Thanks @frankdiox !

Can you also add this to the Changelog.
I think you can also detect iOS version with this plugin so can you add that as well: https://cordova.apache.org/docs/en/3.0.0/cordova/device/device.version.html

frankdiox:
@argelius You mean `isIOS7above` method, right? I just changed it.

argelius:
Yes, testing this in Chrome is hard. You could mock the plugin though. I will merge it anyway. Thanks :)

PR #1006, opened 2015-11-05
fix(ons-notification.alert()): Fixed #999.

### LINKED ISSUE #999: ons.notification.alert looks weird on Android
The generated HTML is incorrect so the message looks weird on Android.

screenshot from 2015-11-04 12 33 40

anatoo:


argelius:
Looks good! Thanks!

PR #1019, opened 2015-11-09
fix(ons-tab): closes #1017

### LINKED ISSUE #1017: Tabbar is broken
In this pattern:
https://github.com/OnsenUI/OnsenUI/blob/master/css-components/www/patterns/details.html

This error is thrown:

```
Uncaught TypeError: tab.setInactive is not a function
```

frankdiox:
Looks like `ons-tabbar` methods are not defined by the time `ons-tab` calls them if `active` attribute is present.


argelius:
Thanks @frankdiox ! Works great.

PR #1025, opened 2015-11-10
feat(ons-back-button): automatically hides if navigator has only 1 page

frankdiox:
This is Angular free so there is no need to use a watcher or `ng-show` anymore. Check if you like this implementation @argelius @anatoo 

Also, with the pointer to page's back-button added in Navigator page objects we can do something to modify back-button's popPage on demand. Something like `navigator.getCurrentPage().addOptionsToBackButton(object)`, so it could have `onTransitionEnd`, `refresh`, etc.


argelius:
Thanks @frankdiox! I think this looks good.

PR #1039, opened 2015-11-13
feat(components): 'var' attribute without Angular

frankdiox:
@argelius I guess we also need `var` attribute when using OnsenUI without Angular. I didn't remove [this code](https://github.com/OnsenUI/OnsenUI/blob/master/framework/services/onsen.js#L456) from `angular-onsenui.js` to avoid possible BC breaks so it overwrites these variables with the old stuff (it always loads after `onsenui.js`). Perhaps we should remove it and only let WebComponents to write in JavaScript global scope and Angular in its own scope?

Also, I added `var` attribute for every documented component that had it with Angular. Should we add it to `ons-fab`, `ons-progress`, etc? And is it necessary on `ons-toolbar` and `ons-toolbar-button` (they have it right now)?

### DIFF
--- a/core/elements/ons-alert-dialog.es6
+++ b/core/elements/ons-alert-dialog.es6
@@ -73,6 +73,10 @@ limitations under the License.
       this._visible = false;
       this._doorLock = new DoorLock();
       this._boundCancel = this._cancel.bind(this);
+
+      if (this.hasAttribute('var')) {
+        ons._defineVar(this.getAttribute('var'), this);
+      }
     }
 
     _compile() {
--- a/core/elements/ons-carousel.es6
+++ b/core/elements/ons-carousel.es6
@@ -124,6 +124,10 @@ limitations under the License.
       this._setupInitialIndex();
 
       this._saveLastState();
+
+      if (this.hasAttribute('var')) {
+        ons._defineVar(this.getAttribute('var'), this);
+      }
     }
 
     _onResize() {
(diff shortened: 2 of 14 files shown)


argelius:
Is the `var` attribute necessary? Both React and Angular have ways to access to components and for jQuery I think it's more natural to just select:

`$('#menu')[0].open()`

What do you think @anatoo?

anatoo:
@argelius I agree your opinion.
I think `var` attribute is not necessary on onsenui.js. Because on Onsen UI v2 all components are just DOM elements and basically other frameworks have a its own way to refer DOM elements.

frankdiox:
Certainly it's easy to access to the components with jQuery, querySelector or anything, I just thought it's a bit weird having them in the global scope only if we use Angular. In any case I guess we need docs to specify that `var` attribute and some other things are only for Angular. Closing this now, thanks!

frankdiox:
@argelius @anatoo I'd like to bring this topic again. I know this is not really necessary anymore but there will be users porting their apps from 1.x to 2.0 just because they want to get rid of AngularJS. For backward compatibility purposes, do you think we should add this as a possibility? We can do as @argelius suggest and attach variables to `ons` object or anything else. We should also inform this is for backward compatibility and it's not recommended if you don't use AngularJS.

anatoo:
I never revive "var" attribute on `onsenui.js`. But I think it's necessary to announce that "var" attribute is no longer available for jQuery user.

argelius:
I agree that it's not necessary so I will close this PR.

PR #1061, opened 2015-11-24
class.js in angular-onsenui.js

frankdiox:
`class.js` is only used by Angular so I think it's better to put it in `angular-onsenui.js` and reduce `onsenui.js` size a bit.


argelius:
Thanks @frankdiox !

PR #1073, opened 2015-11-26
Animation options parser

frankdiox:
@argelius @anatoo This PR is finished, please review it when you have some time.

`animation-options="{delay: 1, duration: 1, timing: 'ease-in'}"` is currently broken because it's not a valid json for `JSON.parse( ... )`. 

This supports things we don't currently use in `animation-options`, such as arrays and nested objects.


core/lib/animationOptionsParser.spec.es6
```diff
@@ -0,0 +1,23 @@
+describe('ons._animationOptionsParser', () => {
+  it('exists', () => {
+    expect(ons._animationOptionsParser).to.be.ok;
+  });
+
+  it('accepts valid strings', () => {
+    expect(ons._animationOptionsParser.parse('{}')).to.be.ok;
+    expect(ons._animationOptionsParser.parse('{ _v: false,  $qw : .1, w$_ : {x : 1e3}, as :\' a\', _2  : [6, {x : \'%^&\'}, 8] ,  q:{x:{p:[1,2]}} , xc:"bnm"   }')).to.be.ok;
+    expect(ons._animationOptionsParser.parse('[]')).to.be.ok;
+    expect(ons._animationOptionsParser.parse('[{ i: 1},3,  "qw",\'zx\', 1e2]')).to.be.ok;
```
argelius: Should this be valid? I think it should only accept objects.

core/lib/animationOptionsParser.spec.es6
```diff
@@ -0,0 +1,23 @@
+describe('ons._animationOptionsParser', () => {
+  it('exists', () => {
+    expect(ons._animationOptionsParser).to.be.ok;
+  });
+
+  it('accepts valid strings', () => {
```
argelius: Can you also write tests that confirm the integrity of the objects?

argelius:
@frankdiox Can you also write tests that confirm that the regression is fixed? I mean, test that the attribute actually works for some element.

frankdiox:
@argelius Done!

argelius:
Thanks @frankdiox !

frankdiox:
@argelius Conflict solved.

PR #1074, opened 2015-11-26
Fix 1048

d3orn:
This changes the behaviour of shouldFillStatusBar to return a promise.
This has to be done if we want to use the device plugin because we have to wait for a deviceready before we can get the platform and version of the device.

Attention: This will break some tests!


d3orn:
looks like the page__status-bar-fill is not correctly added but I hope this provides a start for you guys

argelius:
This looks good. I will review it and merge if there's no problems. I made a new branch and opened a new PR so I will close this one.

https://github.com/OnsenUI/OnsenUI/pull/1081

PR #1080, opened 2015-11-27
fix(css-components): Load custom CSS at end of body to override font …

argelius:


argelius:
@anatoo This solves an issue where the iframes on the CSS components page will be displayed incorrectly because the CSS is loaded in the incorrect order.

I will just merge it since I don't think it will break anything.

anatoo:
:+1:

PR #1081, opened 2015-11-27
fix(*): Fix status bar filling on iOS Cordova apps.

argelius:
This changes the behaviour of shouldFillStatusBar to return a promise.
This has to be done if we want to use the device plugin because we have
to wait for a deviceready before we can get the platform and version of
the device.


argelius:
@d3orn I've merged your fixed and also made some tests. I had to also change som CSS to make the status bar filling work nicely with material tabs and toolbar.

PR #1093, opened 2015-12-03
Material range expand

frankdiox:
Modifying `var-material-range-thumb-vertical-margin` changes the height of the invisible body.

@argelius Please review when you have some time.


argelius:
Thanks @frankdiox ! I tried it and it seems to work well. I think this is a good tradeoff since we can't increase the hitbox in the x-directiono.

PR #1096, opened 2015-12-07
Dialog, Alert-Dialog and Popover fixes

frankdiox:
Some issues addressed here:
* Combining animations + `border-shadow` + `border-radius` is killing for slow devices and crosswalk. This has been fixed by changing DOM structure and splitting those CSS properties into two different elements. This could be [BC].
* Fixed `overflow: hidden` in `ons-dialog` and `ons-popover`. When adding an `ons-page` inside those components borders became squared due to a bug in webkit where `overflow: hidden` doesn't work.
* Removed iOS shadow for `ons-dialog`, `ons-alert-dialog` and `ons-popover` since there is no shadow in the real iOS style.
* Changed background color for `ons-dialog` and `ons-alert-dialog` to fit the original style better.


argelius:
@frankdiox Looks good! Will merge now. :+1:

PR #1101, opened 2015-12-08
fix(ons-back-button): auto hides on 'resetToPage()'. Closes #1100

### LINKED ISSUE #1100: ons-back-button doesn't automatically hide on `resetToPage()`
It also happens with `replacePage()` when there is only 1 page.

frankdiox:
@argelius Please review when you have some time.


argelius:
Thanks @frankdiox !

PR #1105, opened 2015-12-09
Improve list

argelius:
This adds some styles to easier format lists:

```
<ons-list-item>
  <div class="list__item__left">
    <img src="image.png" class="list__item__thumbnail">
  </div>

  <div class="list__item__center">
    <div class="list__item__title">
     Title
    </div>
    <div class="list__item__subtitle">
     Subtitle
    </div>
  </div>

  <div class="list__item__right">
    <ons-switch></ons-switch>
  </div>
</ons-list-item>
```

It uses flexbox with the fallbacks and it doesn't break backwards compatibility since you can still use the list without them.

Using this I was able to remove almost all custom CSS from the Material Design patterns and also a lot of the CSS from the flat design patterns.

@anatoo I will merge this if there are no problems.


argelius:
@anatoo I will merge this if there is no problem.

anatoo:
OK. It looks good to me!

argelius:
Thanks @anatoo !

PR #1106, opened 2015-12-10
740 fix

andipavllo:
@argelius when you have time please take a look at this. I decresed the transaction delay from 0.2s to 0.15s as it seems to be similar to the default ios/android behavior.


css-components/components-src/stylus/components/list.styl
```diff
@@ -256,12 +256,9 @@ list__item()
 .list__item--tappable
   transition background-color 0.2s linear
 
-.list__item--tappable:active
+.list__item--tappable-active
```
argelius: It looks like the modifier is called "tappable-active". Can you change to `.list__item--tappable--active`?

core/elements/ons-list-item.es6
```diff
@@ -55,6 +61,14 @@ limitations under the License.
     _shouldLockOnDrag() {
       return this.hasAttribute('lock-on-drag');
     }
+
+    _onTouch() {
```
argelius: Write test for this method.

core/elements/ons-list-item.es6
```diff
@@ -55,6 +61,14 @@ limitations under the License.
...
+        this.className = this.className + ' ' + 'list__item--tappable-active';
+    }
+
+    _onRelease() {
```
argelius: Write test for this method.

andipavllo:
@argelius should be ok now

argelius:
@andipavllo This breaks the behavior if you're only use the CSS components. Please don't remove the `:active` pseudoclass from the CSS Components.

argelius:
@andipavllo I will merge this today since were actually removing the `tappable` modifier.

andipavllo:
@argelius oh I see, do I still need to fix the CSS? I had no time to do it in the last few days.

argelius:
@andipavllo I have added the "tappable" attribute as we discussed in the meeting. This attribute will work nice on iOS. I restored the "tappable" modifier so there is still an option if you only use the CSS.

PR #1137, opened 2015-12-18
fix(ons-tabbar): Reset style for persistent tabs. Fixes #1132

### LINKED ISSUE #1132: Persistent tabs with animation disappear
When changing persistent tabs in an `ons-tabbar` `style="transform: translate3d(100%, 0px, 0px); display: block; transition: all 0.15s ease-in initial;"` is not removed from `ons-page` styles after the animation ends, thus making it invisible.

frankdiox:
@argelius Persistent tabs maintain transform/opacity changes depending on the animation. Removing `style` attribute works well but do you think it would be a problem for anything else?

It's also possible to just reset the necessary properties:

`element.style.transform = element.style.transition = element.style.display = element.style.opacity = '';`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 CHANGELOG
 ====
 
+v2.0.0-beta.4
+----
+ * ons-tabbar: Fixed [#1132](https://github.com/OnsenUI/OnsenUI/issues/1132).
+
 v2.0.0-beta.3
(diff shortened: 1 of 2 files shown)


argelius:
@frankdiox I think this is OK, even if it might remove styles added by the user. There is no real use-case where styles needs to be added inline to the `<ons-page>` element.

PR #1150, opened 2015-12-25
Carousel perf fix

andipavllo:
@argelius please take a look at this when you have time


argelius:
We will not include this in the current version. Possibly in 2.0.1

PR #1157, opened 2016-01-07
Replace browserify with Rollup.js

argelius:
This PR replaces browserify with Rollup for bundling.

Some improvements:

* File size is smaller
* The code is cleaner.
* More customizable than browserify bundling.

I also replaced all *.es6 files with *.js since *.es6 is not a standard file extension. Now you don't need to customize your editor to get correct source code highlighting.

Rollup.js adds an UMD to `onsenui.js` so it can easily be consumed with e.g. browserify.


argelius:
@anatoo What do you think about this?

I think this bundle is a lot better than the one produced by browserify.

anatoo:
It almost looks good to me.

There is only one thing. I think changes of import path is not needed. You should add 'core/src' to module path.
image

argelius:
Thanks @anatoo! I will restore those those imports

argelius:
@anatoo I have restored the import paths. I wrote a simple plugin for Rollup to make it work since it doesn't support module import paths by default.

If there's no problem I will merge these changes.

anatoo:
@argelius It's great! :+1:

PR #1163, opened 2016-01-08
fix(ons-page): fix status bar fill for iOS and ons-bottom-toolbar

### LINKED ISSUE #1047: BUG:ons-bottom-toolbar In the IOS on the page will go up to the top as v1.3.13
ons-bottom-toolbar In the IOS on the page will go up to the top

danjarvis:
The fix added for [iOS9](https://github.com/OnsenUI/OnsenUI/commit/eec03eeb6624f995ada9e949659c06f1b553b3d1)
causes an issue if an `ons-bottom-toolbar` is being used (either with an
`ons-toolbar` or stand alone).

When an `ons-bottom-toolbar` is being used the CSS selectors for
adjusting the page header to account for the iOS status bar are no
longer applied because an extra `<div />` is added to the DOM.

Fixes #1047.

### DIFF
--- a/core/css/common.css
+++ b/core/css/common.css
@@ -60,7 +60,8 @@ input, textarea {
   background: transparent !important;
 }
 
-.page__status-bar-fill + .page__background + .page__content {
+.page__status-bar-fill + .page__background + .page__content,
+.page__status-bar-fill + .page__bottom-bar-fill + .page__background + .page__content {
   padding-top: 20px;
 }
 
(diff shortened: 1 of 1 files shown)


argelius:
Hi @danjarvis !

I'm unable to reproduce this issue in the current master branch. What version are you using?

danjarvis:
Hey there @argelius !

The issue was occurring in v1.3.14. I see that you just released 1.3.15 so I'll see if I can reproduce it there.

argelius:
I want to merge this but I see you use `~` selectors. There is a bug in the iOS9 Webview that breaks the behavior of the `~` selector so we can't use it. Is it possible for you to implement this using only `+` selectors?

danjarvis:
@argelius Yeah I can make that change. Also... it seems that 1.3.15 is not available through bower.

frankdiox:
@danjarvis 1.3.15 just changes things in the website, OnsenUI itself is exactly the same than in 1.3.14 so we haven't release it in the distribution repo.

danjarvis:
@frankdiox understood -- thank you!

@argelius just updated the PR with your requested changes.

Also, do you have any good information on the iOS9 WebView bug? In searching around it appears that the bug only manifests when updates need to occur based on pseudo-classes: `.foo:hover ~ .bar`, `foo:checked ~ .bar` or if a selector includes multiple class names and one is removed: `foo.bar ~ baz`.

IliaSky:
Thank you for the pull request @danjarvis, but actually we decided to just remove the `.page__bottom-bar-fill` element altogether. It makes no sense to add an element simply for the purpose of changing the styles of it's siblings - adding a class to the parent can serve the same function. Thus around 3 weeks ago we [removed](https://github.com/OnsenUI/OnsenUI/pull/1291) it.

Once again thank you for your help.

danjarvis:
@IliaSky That makes sense. Thanks for the update!

Issue #1175, opened 2016-01-15
Double clicking a menu element pushes the same page twice


IliaSky:
The same behaviour was observed in #868.
Reproducing the double click is not very easy, but not impossible either and having the same page in the page stack twice is surely not the expected behaviour.

In pushPage I was able to find an option cancelIfRunning which supposedly fixes this issue, but even after enabling it I was still able to reproduce the issue by a manual double click.

Since it wasn't really easy to reproduce I made a [codepen example](http://codepen.io/IliaSky/pen/rxGazV?editors=001) and recorded my actions in a 
[repeater](https://itunes.apple.com/us/app/repeater/id443370764?mt=12) [record](https://github.com/OnsenUI/OnsenUI/files/92056/click.record.txt) (.record).

PR #1176, opened 2016-01-15
fix(ons-navigator): fix the cancelIfRunning functionality

IliaSky:
Currently if a user manages to click twice on a menu element even with  `{ cancelIfRunning : true }` he will have the same page pushed twice (Issue #1175).
The reason for that is that _isPushing is being set in a callback function, meaning that the second execution of the click handler can start before the flag is set to true. This flag isn't used in many places so side effects from this should not exist. Both `core-test` and `e2e-test` show no errors after this change.
The original assignment of _isPushing to true is left unchanged due to the function _pushPageDOM being called in bringPageTop.
Replication of the bug is hard to achieve by hand so it can be done with [this codepen](http://codepen.io/IliaSky/pen/jWGWeg?editors=001) and [this repeater script](https://github.com/OnsenUI/OnsenUI/files/91770/click.record.txt) (.record). The codepen also shows the relative part of the fix, so it can be enabled and seen easily.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ v2.0.0-beta.6
  * css-components: Fixed [#1162](https://github.com/OnsenUI/OnsenUI/issues/1162).
  * ons-input: Add "ons-input" component and remove "ons-material-input" component.
  * ons-range: Add "ons-range" component.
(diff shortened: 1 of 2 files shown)


argelius:
Thanks @IliaSky !

Can you please also add this to the Changelog?

argelius:
Looks good! Thanks @IliaSky :)

Issue #1184, opened 2016-01-24
<ons-fab> not animated when leaving tab.


argelius:
When using an `<ons-fab>` element inside a tab, the `<ons-fab>` element is only animated when entering a tab but when leaving the tab it just disappears.

It would look great if it was animated also when leaving the tab so if there are two different `<ons-fab>` elements in the two tabs the button will shrink and then grow.

frankdiox:
@argelius With the current implementation there would be a small side effect. Page's `hide` event would be fired right before the tabbar animation starts rather than at the end. We can always look for `ons-fab` descendants and hide only those elements instead of hiding the whole page before the animation. What do you think it's better?

frankdiox:
Also, currently `ons-fab` is animated when entering only persistent pages that already exist. If it's a new page it will be shown by default without animation. We can also change it to be always hidden by default. Its `show` method is actually triggered when the parent page is initialized (due to the propagation API) so it would work without further modifications.

argelius:
@frankdiox Actually, I tried master branch after the merge but I'm still having the same issue.

frankdiox:
@argelius It works when you use any animation in the tabbar but not with 'none'. The reason is that `animit` is async but when the animation is 'none' it just runs the callback synchronously without calling `animit`. A simple timeout in `TabbarNoneAnimator` fixes it.
With 'none' animation in the tabbar the `ons-fab` of course won't display its hiding animation since the tab is hidden instantly.

IliaSky:
@argelius are you still able reproduce the issue?

PR #1185, opened 2016-01-25
Fab animation

frankdiox:
@argelius This is done. As I mentioned #1184, page's `hide` event is fired slightly before but I think it's OK. The animation looks much nicer now :)


argelius:
Thanks!

PR #1186, opened 2016-01-25
refactor(ons-splitter): Renamed some methods.

frankdiox:
@argelius I think this is better. Sliding menu still has a `isOpened` method but I guess we should not change since it's a BC.


argelius:
I like these names better! Thanks :)

Issue #1187, opened 2016-01-25
Elements compiled twice if they are duplicated by copying HTML string.


argelius:
If an element is attached to the DOM, it will be compiled and maybe gain some children.

E.g., when the `<ons-dialog>` element is attached it gains two children: `.dialog` and `.dialog-mask`. If the element is copied somewhere else by duplicating the `innerHTML` (not by using something like `appendChild`) a new element will be created and the `_compile()` method is executed again, adding two new children.

Of course, this is also true for other elements than `<ons-dialog>`. I just used it as an example.

A lot of frontend frameworks do manipulations like this so we must stop all components from being compiled twice.

argelius:
I think we also need to write tests for all components to verify that they will "survive" an operation like this. This means that we can not add anything in the `createdCallback` which is not encoded in the HTML. 

We can't add properties or event listeners to the elements but we can change style and add elements since that will be properly encoded in the string.

frankdiox:
@argelius There is another problem. What if the copied element is, let's say, an `ons-navigator` that already contains some pages. The new element would have 4 pages in the DOM but its `this._pages` property would be empty. Should we recreate all the internal properties?

frankdiox:
@argelius The DOM shouldn't be modified anymore after the last PR, but as we mentioned before, there are still problems regarding internal variables that will be gone when copied. Most of the issues have an easy fix. For example, `ons-page` saves a boolean variable `_isShown` that could be stored as an HTML attribute instead without troubles. There are some bigger problems, though, with navigator's page stack, lazy-repeat's delegate and templates (and perhaps some others). Should I fix those issues in another PR?

argelius:
Let's discuss `<ons-navigator>` next meeting. I think in this case we might be able to handle it when wrapping it in the framework in question.

I think if we will find other issues while writing Angular 2 and React.js bindings but I think this particular issue solves some of these problems.

PR #1189, opened 2016-01-26
Avoid compiling twice.

frankdiox:
@argelius This maintains the same DOM structure and attributes when an element is copied using `innerHTML`. I added tests to every element, even to those that don't have a real `_compile` method just to be sure the DOM stays the same. 

As mentioned in #1187, internal variables would be a problem. I can make some modifications to minimize the problems.

There is currently a problem with `ons-template` since it removes its content on createdCallback, so the copy and the original one will be empty (thus passing the test).


argelius:
Thanks @frankdiox !

Just checked it. Looks great :+1:

PR #1190, opened 2016-01-27
Page fix

frankdiox:
@argelius I realized that if `ons-page` is the main component, when trying to propagate `show` action its descendants are not defined. Now that `ons-fab` is hidden by default it would not be shown (check `examples/ons-fab/index.html`). This small change fixes it.


argelius:
Thanks @frankdiox ! Looks great :)

PR #1194, opened 2016-01-29
feat(ons-tab): Tabs are always persistent.

frankdiox:
@argelius This is a quick feature that I think could be useful. I find a bit annoying that a persistent tab takes a while to load the first time. This modifies the default behavior to load every persistent tab at the beginning except if they have `lazy-load`. We can do it the other way around and add `early-load` or something similar. Feel free to close the PR if you don't find it useful.


argelius:
Thanks @frankdiox! Actually I've been wanting to make this the default behavior so maybe we can discuss this after I come back. I will not merge it now.

I see no reason why the tabs should not be pre-loaded, it just causes issues like the one you mentioned.

frankdiox:
@argelius Added to the changelog!

argelius:
@frankdiox 

I talked to @philolo1 about this and I think it might be good if we actually remove the `persistent` attribute and make preloading the tabs default behavior.

It will make it possible to make swiping tabs and it also makes it a lot easier to write bindings for frameworks.

What do you think? Maybe you or @philolo1 can change this PR to do this.

frankdiox:
@argelius In that case there is no possibility of lazy-load, rigth? I think that's how it should be, actually. I can do it :)

argelius:
Thank you @frankdiox :heart:

philolo1:
@argelius yes, i think this is a good idea. I think with this change we would not need to do much of DOM manipulation and only do css/style changes, that would make integration much in react much easier.
@frankdiox great!

frankdiox:
@argelius @philolo1 I think it's done!

philolo1:
@frankdiox great, what is the '_remove-element' property used for in the code?

frankdiox:
@philolo1 `loadPage` method creates a non-persistent page that must be removed after changing tabs. That attribute just marks the pages created with `loadPage`. I used the same name than an internal variable, perhaps I should rename it to `_non-persistent`.

frankdiox:
@philolo1  I fixed an old bug in the last commit and made some refactoring. `_remove-element` attribute is not needed anymore.

philolo1:
@frankdiox thanks looks good now, that makes more sense now, not to have both an attribute and an option.

frankdiox:
The previous bug made it compulsory to mark the page with an attribute, otherwise some pages woudln't be removed from the DOM. But it's fixed now :sweat_smile:

argelius:
I just tested this. Works great :+1:

PR #1195, opened 2016-01-29
fix(ons-fab): Fix #1192 - fab disappearing before animation.

### LINKED ISSUE #1192: ons-fab disappear in animations when position is fixed.
`ons-fab` moves out of the screen when the page is changing. This only happens when `position` attribute is set.

IliaSky:
Also probably fixes the scenario in which #1184 was not fixed.

The live version fabs also have collisions with `ons-toolbar` if they are in a corner which has a toolbar - they overlap (even without any animations).

During the page transitions the `transform: translate3d(0,0,0)` style which enables hardware acceleration makes the `ons-page` behave as a container element for the fabs, making them blink before the animations.

With this change it's activated by default, so they don't overlap with the toolbars and stops them from blinking and disappearing at the start of the animation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ CHANGELOG
 v2.0.0-beta.7
 ----
  * core: Fixed [#1181](https://github.com/OnsenUI/OnsenUI/issues/1181).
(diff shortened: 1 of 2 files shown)


argelius:
Thank you @IliaSky! Makes sense to always have this enabled if it's added by the animator anyway. I will merge this.

PR #1198, opened 2016-02-01
feat(ons-ripple): Improve ons-ripple. Closes #1193

### LINKED ISSUE #1193: Improve Ripple effect to make it more similar to the real thing.
:dog2:

IliaSky:
The new version has more interactions based on different touch events like hold, drag, etc. 

It also has a secondary background and some more options.


argelius:
Thanks @IliaSky ! I will make some reviews.

core/src/elements/ons-ripple/index.js
```diff
@@ -0,0 +1,220 @@
...
+
+*/
+
+import util from 'ons/util';
+import BaseElement from 'ons/base-element';
+import GestureDetector from 'ons/gesture-detector';
+import AnimatorJS from './animator-js';
+import AnimatorCSS from './animator-css';
+
+var Animator = [AnimatorJS, AnimatorCSS][1];
```
argelius: Please use `const` here.

core/src/elements/ons-ripple/index.js
```diff
@@ -0,0 +1,220 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+
+*/
+
+import util from 'ons/util';
+import BaseElement from 'ons/base-element';
+import GestureDetector from 'ons/gesture-detector';
+import AnimatorJS from './animator-js';
+import AnimatorCSS from './animator-css';
```
argelius: Seems like this is not used, so importing it will increase the file size.

core/src/elements/ons-ripple/index.js
```diff
@@ -0,0 +1,220 @@
...
+      this.attributeChangedCallback(e, null, this.getAttribute(e));
+    });
+  }
+
+
+  _compile() {
+    ['_wave', '_background'].forEach(e => {
+      this[e] = document.createElement('div');
+      this[e].classList.add('ripple_' + e);
+      this.appendChild(this[e]);
```
argelius: Since `<ons-ripple>` can be used lots of times in the same page (especially in lists) I think we should avoid doing `appendChild` to the DOM. You could try using `document.createDocumentFragment()` instead and attach that.

I don't know how big the performance win is but I believe it's important that the initialization of this element is very performant.

core/src/elements/ons-ripple/index.js
```diff
@@ -0,0 +1,220 @@
...
+      return this._onRelease(e);
+    }
+    if (['left', 'right'].indexOf(e.gesture.direction) != -1) {
+      this._onTap(e);
+    }
+  }
+
+
+  attachedCallback() {
+    if (this.getAttribute('target') != 'autofind' &&
```
argelius: This block is super slow. Try profiling and you will see.

I tried removing it and got a 5x speed increase.

core/src/elements/ons-ripple/index.js
```diff
@@ -0,0 +1,220 @@
...
+      this.setDisabled(true);
+    } else {
+      this._gestureDetector.on('tap', this._onTap);
+      this._gestureDetector.on('hold', this._onHold);
+      this._gestureDetector.on('dragstart', this._onDragStart);
+    }
+  }
+
+
+  detachedCallback() {
```
argelius: This might be true for all elements, but there seems to be an issue in Chrome where the `detachedCallback` is called before `createdCallback` some times. :(

Creating ~100 elements and detaching I managed to get it happen ~30% of the time...

It looks like conditionally initializing `this._gestureDetector` in `attachedCallback` will solve that problem.

I will try to create a simple example and file a bug report to Chromium. There are lots of issues with the implementation in Chrome unfortunately :(

argelius:
@IliaSky The animation looks great! I will do some additional review later.

I hope to make this component performant since there could be an issue if the developer creates a large list with an `<ons-ripple>` element attached to each one.

Maybe you can move some code to the touch event listeners since they will only be called when you touch the element as opposed to doing a lot of work in the initialization phase.

core/src/elements/ons-ripple/index.js
IliaSky: The element used to have two animator libraries which it could use, depending on what is preferred. The API is the same for both. Nevertheless I commented out the one not being used right now.

core/src/elements/ons-ripple/index.js
IliaSky: The check itself is required in order for the component to work. Its slow, because it requires the rendered style of the element, not the inline style of the element.

Moved the check to be done when the element is clicked for the first time. Now there is no delay from the initialisation.

core/src/elements/ons-ripple/index.js
IliaSky: I moved it over there, but the test got a little extra clutter because of that.

core/src/elements/ons-ripple/index.js
IliaSky: Document fragments increase performance mostly because they make all changes to the DOM at once, hence making the browser do only one reflow instead of multiple ones. 

However if the modifications are made outside the DOM, using document fragment actually slows down the process - [jsperf test 1](http://jsperf.com/createdocumentfragment-vs-appendchild-on-non-appended-e/32) . In our web components, as far as I understand - _compile is executed in createdCallback, not in attachedCallback, so the actions here are before the element is attached to the DOM.

In this case also even if it was inside the DOM I am appending only 2 children - again at least in the current browsers creating a fragment for 2 children does not seem to be beneficial, and again makes it slower - [jsperf test 2](http://jsperf.com/docfragment-test/14).

Therefore I have not modified this part in the recent commit.

In the second jsperf I have also added another case if the code for the two elements is copy-pasted. There seems to be around 3% performance improvement if its done. There are probably also other places where I could make the same change if we want to squeeze out the best performance. If you think this is necessary please tell me and I will fix them.

IliaSky:
After applying the suggested changes it seems that the GestureDetector seems to be the main bottleneck. I am not sure if I should try to use it in a way which tries to share the instance across all `ons-ripple` elements or if it would be better to try to make that optimisation it from the GestureDetector class itself.

argelius:
Thanks @IliaSky ! It looks like the initialization is a lot faster now.

I agree that `GestureDetector` is the main bottleneck. There is a large memory overhead since it has to create an instance for each `ons-ripple` element.

I tried on my phone with ~100 ripple without problem. When I approach ~300 elements the animations starts to stutter but it's still working. We might revisit this element later to make it faster.

argelius:
@IliaSky By the way, I replaced some comparisons (`==`) with strict comparison (`===`) since JavaScript does some strange conversions when you use `==`.

argelius:
I will merge this now. Great job!

PR #1199, opened 2016-02-02
Auto Styling feature. Extended ons-input.

### LINKED ISSUE #1125: Support multi-platform design
Modify/add components, attributes and modifiers to automatically adjust the design based on the running operating system.

* By default, `modifier` attribute is automatically set based on the running OS.
* New `ons.disableAutomaticStyling()` function to prohibit automatic platform adjustment.
* New `<ons-button>`, `<ons-range>` and `<ons-input>` components.
* Remove `<ons-material-input>` component.
* New `effect` attribute to manually enable/disable ripple effect (default: `ripple` for Android).
* New `disable-auto-styling` attribute to manually enable/disable modifier and ripple effect (default: `false`)
* Change `<ons-list-item>` to support ripple effect.
 * Also change `tappable` from modifier to an attribute.
 * Also remove effect from `chevron` modifier.
* Keep `<ons-ripple>` as it is.
* Add new modifier to `<ons-fav>` to remove show when displaying in iOS.
* Add `<ons-if-platform>` element.
 * Example usage: `<ons-page>...<ons-if-platform android>...</ons-if-platform>...</ons-page>`

frankdiox:
@argelius Automatic styling is almost done. Closes #1125 . This adds `ons.disableAutoStyling()` and `ons.enableAutoStyling()` in order to map iOS styles to MD and the other way around. It works for components created after those method are called, so the first one should be called when the app starts. Also adds `auto-styling-disabled` attribute to disable the feature in specific components.

Please give me some feedback specially about this commit 06c142c . Some of the changes in that commit are:

* Since our checkbox and radio buttons were only CSS classes there is no way to auto style them. I have extended the new `ons-input` to add `type="checkbox"` and `type="radio"`, right as the normal HTML input element. Those three types are quite similar and have the same DOM structure with different classes. The checkmark for checkbox and radio is the label in text/password.

* `label` attribute has been renamed to `placeholder` so it is also displayed on iOS without effect.

* I have slightly changed the structure of `ons-input` in order to allow "labels" in this way: `<ons-input type="... any ..." > When clicking this text the input will be focused </ons-input>`. That text is displayed on the right side by default or on the left if `content-left` attribute is present.

* It maps `ons-input`'s ID to the inner input as `id="inner-${ parentID }"` so it's possible to add `<label for="inner-...">`

Other things:

* For material, ripple effect is added automatically to `ons-button`, `ons-toolbar-button`, `ons-back-button`, `ons-tab`, `ons-fab`, `ons-speed-dial-item`. Also to `ons-list-item` only if it was tappable in iOS.

* For iOS all ripple effects are removed. If `ons-list-item` had ripple effect it becomes tappable.

* `ons-list-item` that contain both `ons-input` and ripple effect wrap its content inside a `<label>` element to make the input and the rest of the content clickable. Otherwise the ripple effect takes the click (since they would be sibling elements) and the input wouldn't be checked.

* Large buttons in iOS maintain `large` modifier in Android (it only adds `width: 100%`) to avoid destroying the layout making the button smaller.

Suggestions are welcome :)


masahirotanaka:
Thanks. Great work @frankdiox!

I like all of your decisions mentioned above.
I just would like to know if you also will work on the items we discussed earlier, like adding `effect` attribute. I noted them down on the URL below, so please also take a look at it and see if you want to include some of the enhancements.

https://github.com/OnsenUI/OnsenUI/issues/1125

frankdiox:
@masahirotanaka Thanks!
Yes yes, I implemented all of this considering that issue, so I'm using `effect` attribute. I was thinking that maybe we could make `tappable` also a type of effect instead of an attribute since I guess nobody will want tappable + ripple.
I will also try to make a new MD default animation for `ons-navigator` since the current one is for iOS.

frankdiox:
@argelius @masahirotanaka I think this is finished now unless you want to modify or add something else.

I added two new navigator animators that are a bit more similar to MD ones. I'm not specially good with `animit` so please check if you like them.
I also added `ons-if` component as we mentioned in #1125 . `platform` attribute is only usable at the beginning (createdCallback) but `orientation` can be updated. I guess at some point we could add `ons-if plugin=" "` or something like that?

Anyway, I will wait for your feedback before making any further change :smile:

frankdiox:
Should `ons-tabbar` have `position="top"` by default on Android with the auto styling? I think it would be nice but there is a problem: `ons-toolbar` would be displayed below `ons-tabbar` if they include it inside the pages and not as a sibling of the tabbar...

frankdiox:
@argelius @masahirotanaka This is finished. In the end we are using `ripple` attribute instead of `effect="ripple"` as we discussed. Also, this adds `position="auto"` for the tabbar so the position depends on the platform.

masahirotanaka:
Thanks :+1:

argelius:
@frankdiox Great! I will merge this :+1:

PR #1200, opened 2016-02-03
refactor(animationOptionsParser): Only exports \'parse\' method.

frankdiox:
@argelius I have a better understanding of ES6 modules now so I'm changing this a bit.


argelius:
Thanks @frankdiox !

PR #1203, opened 2016-02-04
fix(ons-tabbar): NoneAnimator should be async.

frankdiox:
@argelius With this,  `ons-fab` will display its show animation when tabbar's loadPage animation is `none`. Related to #1184


argelius:
Looks great @frankdiox ! :+1:

PR #1205, opened 2016-02-04
Ons switch draggable

### LINKED ISSUE #894: Draggable <ons-switch>
Make `<ons-switch>` draggable.

Example:
https://www.polymer-project.org/0.5/components/paper-toggle-button/demo.html

IliaSky:
Closes #894.
Also fixes a small bug in `ons.util.createFragment`. The live version ignores all children after the first one, but in the only place where it's being used the code is trying to iterate over them afterwards.


argelius:
@IliaSky Is this implementation finished? In that case I will test it and merge it

IliaSky:
It should be finished, but I have a test failing in CircleCI. I fixed it locally, but on CircleCI it still fails.
Maybe there is a missing gulp build in css-components, which is making the problem over there, but I'm not 100% sure.

argelius:
Thanks @IliaSky ! Works great.

I changed the ripple effect to a box shadow so it doesn't cover the thumb element. I also added a 100ms transition to make it feel nicer.

Actually, I don't really like that it's not activated on hold. It doesn't feel responsive. The native switch will show the ripple as soon as you press the switch. Polymer also does this:

https://elements.polymer-project.org/elements/paper-toggle-button?view=demo:demo/index.html&active=paper-toggle-button

I will merge this, but when you have time can you please see if you can improve this.

PR #1213, opened 2016-02-08
First attempt to make ons-navigator react compatible

philolo1:
@argelius Thisi s my first attempt to start integrating react into Onsenui. It creates a component OnsNavigator and OnsPage that handles push/pop and insert right now. It is tricky because of the dom manipulations you have to tell react about. Please have a look also at examples/react.


philolo1:
i used the package "npm install --dev-save gulp-babel", where should i add this?

ggn06awu:
Is this work expected to work with Onsen 2?

philolo1:
@ggn06awu this is still in the developer state and this commit is only for general discussion how it will/ could look like, but esentially we will support react in OnsenUI 2.0. We will keep you updated!

philolo1:
closing this pull request for now, but have made significant progress and will keep you updated

PR #1219, opened 2016-02-10
perf([ons-list-item, ons-ripple]): Add a gesture detector to the body…

IliaSky:
… instead of each element. This results in a significant performance boost in the initialisation of these elements.


argelius:
Thanks @IliaSky !

Is this PR finished?

IliaSky:
yes

argelius:
This is great! :+1:

Issue #1222, opened 2016-02-10
ons-split-view slides back after scrolling within the content of it


rockneverdies55:
I have an `ons-split-view` with a long list of chevron (tappable) items wrapped in a `ons-splitter-side`. If I scroll the list within the split to select an item, the split view slides back without letting me click on a list item. 
Is there a way to stop this?

IliaSky:
Thank you for the report. This should be fixed when we release beta 7.
In the meantime you can use this ugly hack:
```javascript
onsSplitter.addEventListener('dragstart', function(e){
	if (e.gesture.direction === 'up' || e.gesture.direction === 'down') {
		onsSplitterSide._getModeStrategy()._ignoreDrag = true;
	}
});
```
`onsSplitter` and `onsSplitterSide` being the `ons-splitter` and `ons-splitter-side` elements.

rockneverdies55:
Thanks for taking care of this and offering a working temp solution.

IliaSky:
You're welcome. The fix is already in place, so feel free to use the newer version when it becomes available. I am closing this issue.

PR #1226, opened 2016-02-12
fix(ons-splitter-side): Fix splitter side scrolling issue #1222

IliaSky:
Currently scrolling inside ons-splitter-side on a mobile device closes the side menu after the user has finished scrolling.


argelius:
Thanks @IliaSky ! Looks good.

PR #1229, opened 2016-02-15
feat(ons-switch): Improve draggability, fix #1228, remove auto generated name.

### LINKED ISSUE #1228: CSS Switch is broken
The switch element doesn't work anymore. It only works when using the `<ons-switch>` element but the CSS should work independent of the JavaScript.

Also, `checked` and `disabled` attributes stopped working.

@IliaSky Can you fix this on Monday?

IliaSky:


argelius:
Thanks @IliaSky ! Works great now.

PR #1230, opened 2016-02-15
Notification promises

frankdiox:
@argelius This is done. Notifications resolve when the buttons are clicked to the `buttonIndex` or the `inputValue` in case of prompt.
It also adds a `options.id` parameter that will set the id for the new `ons-alert-dialog`.


argelius:
Thanks @frankdiox ! :+1:

PR #1231, opened 2016-02-15
Page html parameter

### LINKED ISSUE #1208: pageHTML option for Navigator methods.
@argelius @anatoo Currently, as #1207 shows, there is an undocumented parameter `options.pageHTML` available for `pushPage` but not for the rest of navigator's method. Actually, if that option is included, `page` parameter will be ignored but it's still necessary to pass it.

Since `pageHTML` could be a useful option in some cases, like inserting data before creating the actual page, I suggest to modify navigator methods to make them similar to `ons.notification.alert` in #1127. They could maintain the same parameters that right now accept and extend them to be able to get only the option object without any string. The object could contain `page` (url) and also `pageHTML`.

frankdiox:
@argelius This is done. Closes #1208 . It adds `options.pageHTML` parameter for every method of `ons-navigator`. It also makes `page` parameter optional (and adds `options.page` as well). Could you check the Japanese docs? I don't know either how to specify in the docs that both `page` and `options` are optional but at least one of them should be specified. Is possible to have two `@signature`?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ v2.0.0-beta.7
  * ons-tabbar: Correctly applies animation-options.
  * ons-popover: Correctly applies animation-options.
  * ons-alert-dialog: Correctly applies animation-options.
- * ons-navigator: ons-navigator: add "pageHTML" option to "insertPage()" method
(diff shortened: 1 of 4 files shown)


argelius:
Thanks @frankdiox ! :+1:

Issue #1179, opened 2016-01-22
Feature request: Either add ons-scrollbar element or modify ons-scroller.


IliaSky:
The current `ons-scroller` element doesn't seem to do much. 
Even though it has `overflow: scroll` it doesn't provide a scrollbar. In some cases where a user has scrollable content not having a scrollbar could be an issue.
The two options which come to mind to solve this problem are:
1. Addition of a new `ons-scrollbar` element
2. Modification of `ons-scroller`, so that it can have a scrollbar
In either case the final result could look [something like this](http://codepen.io/IliaSky/pen/obowmv).

__Option 1__:
Whenever an ons-scrollbar element is attached to the DOM it makes sure its parent can support it by changing its overflow and position css properties if needed;

__Option 2__:
Add the scrollbar logic to the already existing `ons-scroller`. In that case it can just have an attribute like `scrollbar="visible|hidden|auto|autohide"`.

In both cases it the element could have some different options like touchable radius around it or disappearing some time after a scroll event.

The `ons-scrollbar` has the bonus of being available for non-ons elements, but has a higher chance of causing some issues. The `ons-scroller` option sounds safer, adds meaning behind the tag, but is more limiting.

Which of the two options would be better in your opinion? 
@anatoo @frankdiox @argelius @andipavllo

I should be able to implement it without causing any issues.

frankdiox:
@IliaSky I like it, but since the scrollbar should also be available for `ons-page` (and maybe any other div) perhaps it's better to have it as a new component (option 1). Since it's already listening for `scroll` perhaps it could also throw a custom event when it's near the bottom or some position configurable by the user.

argelius:
I like option 1! :)

Issue #1165, opened 2016-01-13
Implement infinite scrolling on `<ons-page>`


argelius:
Implement a way to execute some code when scrolling the `<ons-page>` element.

```html
<ons-page id="my-page">
  Lots of content.
</ons-page>
```

```javascript
var el = document.querySelector('#my-page');
el.onInfiniteScroll = function(done) {
  // Load data and call done().
};
```

Also implement a directive for `AngularJS` that uses this feature.

anatoo:
:+1:

PR #1232, opened 2016-02-16
Ons scrollbar implementation

IliaSky:
`ons-scrollbar` element implementation #1179
Current iteration supports the following attributes and methods:
__Attributes__:
- `height` - fixes the height of the scrollbar to the specified value. Otherwise it adapts based on the content size and visible part.
- `draggable` - makes the scrollbar draggable.
- `autohide` - makes it disappear after some time after scrolling.
- `hidden` - hides it completely.
- `update-on-scroll` - updates the size and position on each scroll. 
The last setting exists because the content may change without the user having much control over/knowledge of it.

__Methods__:
- `updateScrollbar` - updates the scrollbar height & position.

__Other features__:
Contains the prototype for infinite scrolling #1165:

```
scrollbar.onInfiniteScroll = function(done) {
  // Load data and call done().
};
```

__Additional options which could be implemented__:
- Making the scrollbar area (from top to bottom) clickable so that it can jump to the specified location (could be kind of strange on phones)
- Adding some optional arrows with `jump to home` and `jump to end` functionality during scrolling in a certain direction.
- The current version provides only a vertical scrollbar - I'm not sure if we want to implement the horizontal version if the near future - doesn't seem like something very necessary for the time being.
- Possibly add the option of adding the overscroll shadow effect which exists in native android apps when you try to overscroll up or down. I think iOS has something similar, but I think we get it by default from the browser, though I could be wrong.
- Maybe some other settings? (for example autohide timeout duration or something of similar nature)


argelius:
@IliaSky I just tested this. Works really well.

I added so it runs `updateScrollbar` also when the window is resized. On a mobile device this will only happen when the orientation is changed or when the keyboard is shown so I don't think there will be any performance issues.

argelius:
Thanks @IliaSky !

Issue #1233, opened 2016-02-17
Add e2e test for navigator.insertPage


philolo1:
Looking through the e2e tests I realized the navigator test is testing replacePage and resetPage, push and pop page, but not insertPage. @argelius does it make sense to add one, i can do it.

argelius:
I think it makes sense to add one.

philolo1:
@argelius ok, will do it then, i assigned it to myself, since i am working on the tests right now anyway.

argelius:
Thanks!

philolo1:
@argelius added the test in a pull request.

PR #1234, opened 2016-02-17
test(ons-navigator) add test for "insertPage" function #1233

philolo1:
This adds a test for the "insertPage" function in ons-navigator, just resolving issue  #1233. The test tests both inserting a page at the front of the stack frame and just before the last element of the stack frame.


argelius:
Thanks @philolo1 ! Looks great :+1:

Issue #1035, opened 2015-11-11
Bug: lazy-repeat with ons-scroller


d3orn:
Although lazy-repeat is allowed inside an ons-scroller I can only scroll a couple of elements on non iPhone devices (tested on browser and android). As soon as I remove the ons-scroller it works like a charm.

Keep up the awesome work, cannot wait for the onsen 2 release xD

argelius:
Thanks @d3orn ! We will investigate this.

argelius:
@andipavllo Can you investigate this and release a patch if you solve it?

andipavllo:
@argelius sure, I'll take a look at it today

andipavllo:
@d3orn Are you using the lazy-repeat with an ons-list? If yes, is there any specific reason to add the list inside ons-scroller? It will already scroll by itself.

d3orn:
I guess that was a fix for windows but as far as I can see at the point of writing it works in windows without the scroller so I will just go without it.

IliaSky:
The issue has been fixed, even though we are planning on removing `ons-scroller` soon.

Issue #1029, opened 2015-11-11
ons-lazy-repeat taking full screen height


eidorianyeo:
If i remove ons-lazy-repeat from ons-list-item, this issue won't happen. Basically whenever I use ons-lazy-repeat, it takes on full height, overlap and cover on top of ons-toobar which is at the top of screen. Hope this can be fixed.

argelius:
Hi @eidorianyeo ! Thanks for the report.

Can you provide some code that reproduces this issue?

eidorianyeo:
Hi @argelius ,

Screen shot 1 show alignment without ons-lazy-repeat.
Screen shot 2 show alignment with ons-lazy-repeat.

All I did was add/remove that single function.

![issue1](https://cloud.githubusercontent.com/assets/8005098/11091539/435cb67e-88b6-11e5-9349-ea43167c90ff.png)
![issue2](https://cloud.githubusercontent.com/assets/8005098/11091540/4484d32e-88b6-11e5-8d60-9d7851adc86d.png)

<ons-page ng-controller="testCtrl">
    <ons-toolbar>
        <div class="left"><ons-back-button ng-click="menu.setMenuPage('menu.html');">Back</ons-back-button></div>
        <div class="right">
            <ons-toolbar-button ng-click="menu.toggle()">
                <ons-icon icon="ion-navicon" size="28px" fixed-width="false"></ons-icon>
            </ons-toolbar-button>
        </div>
        <div class="center">Test App</div>
    </ons-toolbar>

    <div class="navigation-bar">
        <div class="navigation-bar__center">
            <input type="search" class="search-input" style="width: 96%; margin: 6px auto 6px auto;" placeholder="Search">
        </div>
    </div>

    <ons-list-header>Ons Header</ons-list-header>
    <ons-list>
        <ons-list-item class="item" modifier="chevron" ons-lazy-repeat="MyDelegate" ng-click="app.navi.pushPage('chatroom.html'); menu.setMenuPage('menuChatRoom.html');">
            <ons-row>
                <ons-col width="80px">
                    <img ng-src="http://lorempixel.com/60/60/?22" class="item-thum"/>
                </ons-col>
                <ons-col>
                    <header>
                        <span class="item-title">{{ item.name }}</span>
                        <span class="item-label">{{ item.category }}</span>
                    </header>
                    <div ng-show="item.description==''">
[...]

Issue #470, opened 2015-03-04
ons-lazy-repeat feature with absolute position


nakomi:
I really like this feature but I just figure out they are positioning with absolute position, so it can only be placed on very top of the page and without some flexibility.

Is there anyway to add padding or container height like ons-scroller did so it can be greater function and maximise the flexibility of UI Design.

Thanks!

argelius:
Right now there's no good way to do this because of the absolute positioning. You can try using `position: relative` for the container though if you want some content above the list.

nakomi:
Thanks!! @argelius 

I will try find some css trick to solve it, but it will be good if we can have it in future.

argelius:
Thanks! We'll try to think of some way to make it easier to use.

I'm reopening this if that's OK.

nakomi:
Thanks a lot!

PR #1241, opened 2016-02-19
fix(ons-lazy-repeat): Fix #1236, #1029, #470 and #1035

### LINKED ISSUE #1236: ons-list-item + ons-lazy-repeat positioned above previous elements
```ons-list-item``` is displaying above previous elements when using with ```ons-lazy-repeat```.

The following image shows the problem
ons-list-item

Also happens with ons-list-header for example
ons-list-item2

This on v2 beta.6

IliaSky:
The first 3 issues are basically the same, the fourth has also been fixed in this commit, although ons-scroller will probably be removed soon anyway.


argelius:
Thanks @IliaSky ! Looks great :)

PR #1242, opened 2016-02-19
feat(ons-carousel): Add `centered` attribute.

IliaSky:
[Some users](http://stackoverflow.com/questions/35463152/onsen-carousel-show-current-item-in-centre) seem to be in need of this.

It would make sense to have this if we support `item-width` and `item-height` properties I guess.
Although even these 2 properties don't seem to be tested very well as the way in which the implementation checks the current item it's impossible to select for example the last one if the size is less than 100%. 

Anyway I don't think this would cause issues and could be useful.


core/src/elements/ons-carousel.js
```diff
@@ -206,6 +238,13 @@ class CarouselElement extends BaseElement {
    */
 
   /**
+   * @attribute centered
+   * @description
+   *   [en]If this attribute is set the carousel then the selected item will be in the center of the carousel instead of the beginning. Useful only when the items are smaller than the carousel. [/en]
```
argelius: This sentence is a bit strange.

core/src/elements/ons-carousel.js
```diff
@@ -665,6 +706,38 @@ class CarouselElement extends BaseElement {
   }
 
   /**
+   * @method setCentered
+   * @signature setCentered(centered)
+   * @param {Boolean} centered
+   *   [en]If true the carousel will be centered.[/en]
+   *   [ja]centeredできるかどうかを指定します。[/ja]
```
argelius: I think this translation is a bit off.

argelius:
@IliaSky Can you ask someone to check the Japanese translations? I think they don't say the same as the English docs.

IliaSky:
@anatoo sorry to disturb you but could you help me with the Japanese documentation?
The lines are `[244, 713, 731, 734]`.

IliaSky:
I updated the translations, so maybe we can merge it now if you think they are ok. :)

PR #1246, opened 2016-02-22
Splitter fix

frankdiox:
@argelius I started a small sample app this weekend and had some troubles with the version in master branch. This PR fixes few bugs in `ons-splitter`:

* `util.arrayFrom(fragment.children)` is empty so there were no `show` events.
* I think `this.innerHTML = ''` should be more efficient than `this.removeChild(this.childNodes[0])` in a loop.
* If the side is not specified it appears on the left but loses some functionality. Set `left` as the default side.
* In `ons-splitter-mask` we try to close rightMenu and leftMenu, but if any of those is missing it will throw an error. I've changed that error to be a rejected promise (it should have been like this since the beginning) and ignore the error with `.catch(() => {})`.

The second commit fixes some console error with tabs.


argelius:
Thanks @frankdiox ! Looks good :)

PR #1247, opened 2016-02-22
feat(ons.notification): Closes #1127

### LINKED ISSUE #1127: New syntax for `ons.notification.alert`
What about allowing the user to just put a string as argument of `ons.notification.alert` ?

```
ons.notification.alert('Hello world!');
```

would be the same as writing

```
ons.notification.alert({message: 'Hello world!'});
```

@anatoo What do you think about this?

frankdiox:
@argelius


argelius:
Great :+1:

PR #1252, opened 2016-02-23
chore(ons-tabbar): add example to use ons-navigator outside a ons-tabbar

philolo1:
This is a simple example where a navigator is outside of a tabbar, a common case in iOS applications.


argelius:
Thanks @philolo1 !

PR #1263, opened 2016-02-26
feat(ons-popover): add cancel event

philolo1:
Currently ons-dialog and ons-alert-dialog throw an 'cancel' event if the user clicks outside of the box. But for ons-popover this is not the case. This fix just makes it more consistent by also adding this event to ons-popover.


argelius:
Thanks!

PR #1264, opened 2016-02-26
fix(ons-list-item): Adds 'center' class.

frankdiox:
@argelius . If the user specifies a center element it will have `.center` class so I think we should add it as well when they don't specify it. It could be useful for selectors.


argelius:
I agree :)

PR #1265, opened 2016-02-26
Fix ripple

argelius:
@frankdiox This fixes the issue with the ripple. I have put `pointer-events: none` on the `.ripple` class and it's children so it shouldn't be able to block any clicks.

Actually while writing this message I noticed a small bug in the ripple so I will fix that as well :)


argelius:
This removes the `target` attribute and attaches event listeners to parent instead of the element itself.

It also fixes a bug where the `_onRelease` method was not called if the user drags outside the element while the animation is running. I added listeners for `mouseout` and `touchleave` to cover all cases.

argelius:
@frankdiox You can merge this if it's OK

PR #1266, opened 2016-02-26
Minor fixes

frankdiox:
@argelius Please merge this before releasing the new beta. Do you remember the memo app crashed silently on iOS? The first commit should fix that. The second commit just puts the ons-fab over the tabbar and other components.


frankdiox:
@argelius `swipe-target-width="50px"` would make the splitter to open only if dragged from the border but once it's open I think the expected behavior should be ignore that attribute.

argelius:
Thanks @frankdiox ! :)

PR #1269, opened 2016-02-29
Css fix

frankdiox:
@argelius Radio button is 1px cropped on the top. This should fix it. Also it includes a fix for `ons-input` to apply width properly.


argelius:
Thanks!

PR #1271, opened 2016-02-29
refactor(ons-speed-dial): extract the class operations to a function '_updateClasses'

philolo1:
This refactors the class ons-speed-dial and exposes a function called updateClasses. This is useful for the react binding: after the DOM is created, react can call this function and all the classes will be set depending on the properties.


philolo1:
@argelius let me know what you think about this approach.

philolo1:
@argelius should i also merge this into  `nav_mod` ?

argelius:
@philolo1 Please wait, I will take a look at it today.

argelius:
@philolo1 I guess we don't need this anymore so I will close it.

philolo1:
yeah

PR #1272, opened 2016-02-29
feat(ons-navigator): Added default  property. CancelIfRunning is now …

frankdiox:
…default behavior.

@argelius Please review this when you have some time.


argelius:
@frankdiox Looks good! Can you also add it to the changelog?

frankdiox:
@argelius Done! I also modified the setter to apply default values when the user change the whole options object.

argelius:
Thanks @frankdiox ! :+1:

PR #1277, opened 2016-03-01
docs(ons-lazy-repeat): update ons-lazy-repeat example to make it work

philolo1:
@argelius in the example the id was missing, so the querySelector would be always null.


argelius:
Thanks @philolo1 ! :+1:

PR #1278, opened 2016-03-02
Fixing ios

frankdiox:
@argelius Please check this small fixes:

* Lift animation translates the leavePage 10% to the top and adds a black mask behind. On iOS the translation of leavePage happens before the translation of enterPage so it's possible to see the black mask for a moment. Adding a delay slightly modifies the animation in every platform so I think this fix is maybe better.

* ons.notification was crashing on iOS when trying call `dialogElement.show` internally because it was not a function yet.

* Since non active tabs are now loaded since the beginning it's necessary to add `display: none` to the element. I think this could be fixed also with `CustomElements.upgrade`, but I guess it's enough with this fix.


argelius:
Thanks @frankdiox ! :+1:

PR #1281, opened 2016-03-02
feat(ons-lazy-repeat): Add delegate render to ons-lazy-repeat

philolo1:
@argelius this will add a delegate render, that can be used to redo the rendering of the items. I am thinking, probably a delegate is not the right way, because i only want to use this in the react binding, what do you think, maybe its better to use a rewritable?


argelius:
@philolo1 I will close this now since we merged it to `nav_mod`.

philolo1:
yeah makes sense

PR #1284, opened 2016-03-03
My nav react

argelius:
@anatoo @patrickklitzke @frankdiox 

I don't want to merge this, I just want some opinions on this approach.

I tried to reimplement the `pushPage()` and `popPage()` methods to make it easier to integrate with React.js. I've removed the `_pages` property since that adds state that will be destroyed if the component is rerendered in React. Instead, the page stack should be defined by the children:

```html
<ons-navigator>
  <ons-page></ons-page>
  <ons-page></ons-page>
  ...
</ons-navigator>
```

This is the new `pushPage()` method. My thought is that the bindings will call `_pushPage` and supply a list, an item to push to the list and a callback function.

The callback function is called when we need the page to be added to the DOM. Onsen UI will push the `page` object to the `pages` list but is actually completely agnostic about what those objects are. We assume that the framework will create and attach the element inside the `update` method and return a promise that resolves when the element is properly attached. After that it will run the animation.

```javascript
  // This is the normal pushPage that takes a string as an argument.
  pushPage(page, options = {}) {
    if (typeof page === 'object' && page !== null) {
      options = page;
    } else {
      options.page = page;
    }

    const run = templateHTML => {
      const element = this._createPageElement(templateHTML);
      CustomElements.upgrade(element);

      return new Promise((resolve) => {
        rewritables.link(this, element, options, element => {
          this.appendChild(element);
          // It calls _pushPage() internally.
          resolve(this._pushPage(options));
        });
      });
    };

    if (options.pageHTML) {
      return run(options.pageHTML);
    }
    else {
      return internal.getPageHTMLAsync(page).then(run);
    }
  }

  _pushPage(options = {}, update = () => Promise.resolve(), pages = [], page = {}) {
    if (typeof options !== 'object') {
      throw new Error('options must be an object. You supplied ' + options);
    }

    options = util.extend({}, this.options || {}, options);

    options.animationOptions = util.extend(
      options.animationOptions || {},
      AnimatorFactory.parseAnimationOptionsString(this.getAttribute('animation-options'))
    );

    const animator = this._animatorFactory.newAnimator(options);

    // Push 'page' to 'pages'. We don't care what these things are as long as `push` method exists.
    pages.push(page);

    // Now we ask the framework to update the DOM.
    return update(pages, this)
      .then(() => {
        return new Promise((resolve) => {
          if (this.pages.length > 1) {
            const l = this.pages.length;
            // At this point the DOM is updated so we run the animation.
            animator.push(this.pages[l - 1], this.pages[l - 2], resolve);
          }
        });
      });
  }
```

I have removed things like events and some options since we can easily add them. `popPage()` is implemented in a similar way.

To illustrate how this works in React:

```javascript
class OnsNavigator extends React.Component {
  constructor(props) {
    super(props);
    
    this.state = {                                                                                    
      pages: []
    };
  }

  update(pages) {
    return new Promise((resolve) => {
      this.setState({pages: pages}, resolve);
    });
  }

  pushPage(comp, options = {}) {
    this.refs.navi._pushPage(options, this.update.bind(this), this.state.pages, comp);                
  }
    
  popPage(options = {}) {
    this.refs.navi._popPage(options, this.update.bind(this), this.state.pages);                       
  }                                                                                                   
    
  componentWillMount() {
    if (typeof this.props.children !== 'undefined') {
      this.setState({pages: React.Children.toArray(this.props.children)});
    }
  }
  
  componentDidMount() {
    this.refs.navi.popPage = this.popPage.bind(this);                                                 
  }   
    
  render() {                                                                                          
    return (
      <ons-navigator {...this.props} ref="navi">                                                      
        {this.state.pages}
      </ons-navigator>
    );
  } 
}
```

So React is handling the page stack but uses Onsen UI for both the animations and the logic. We resolve the promise inside the callback of `setState` since we are sure that the element is added to the DOM at this point.

Please tell me if you think this is a stupid approach :)


frankdiox:
@argelius It looks like a good approach but there is something I'm not sure if I'm understanding. In this case React is not using `rewritables.link`, right? More like skipping `pushPage` and calling `_pushPage` directly. In that case we would need to implement all the options checks and `pageHTML` stuff also in React again.

Isn't it possible to move all the troublesome code inside `rewritables.link`? For example, if `link` receives the necessary data to create the element and also to append it to the DOM. In this case it could get `templateHTML` instead of `element`, but I'm not sure if React could use `templateHTML` to create a page component. Is that possible?

argelius:
@frankdiox Thanks for the feedback!

`rewritables.link` doesn't make sense in React, neither does `pageHTML`. React deals with JSX, not HTML. All frameworks we are targeting have their own templating systems so I think we can just leverage those.

I removed a bunch of code to make it more simple. Things like  triggering events can be put in `_pushPage()` so they happen regardless of the framework used.

We can hook up the Angular bindings to use this instead of `rewritables.link` as well so I don't know if the link function is necessary at all.

One thing I think is nice with this approach is that we can initialize the navigator with any number of pages which can make routing possible and also server side rendering (haven't tried so I'm not sure).

PR #1285, opened 2016-03-04
Lazy repeat refactor and auto calculate feature

IliaSky:
The auto calculate feature is turned on only if the user didn't specify the height.

Currently it works only when all items are with the same height, although I think that's the most common use case. Also the code should be faster if we actually know that since otherwise we need to do more calculations when the user scrolls in order for us to know which items to display.

This pull request also enables es6 in angular 1 bindings.


argelius:
@IliaSky This PR is stale. Can you resolve the conflicts?

IliaSky:
ok

IliaSky:
BTW it's ready for merging :)

argelius:
Thanks @IliaSky !

PR #1291, opened 2016-03-10
Fix bottom toolbar

IliaSky:
The bottom toolbar wasn't working properly in 2 scenarios:
1. when there is no angular
2. in some cases when there is a status-bar-fill (IOS)

This pull request fixes those and also removes the need of  the element `.page__bottom-bar-fill`. This also simplifies the logic in `tryToFIllStatusbar` in `ons-page` as it does not need to work around it.

It's probably going to be easier if we don't add a `.page__status-bar-fill` element too, as it's used only for styling the elements after it and thus can be easily replaced by putting an attribute or class to the parent instead, but since the same logic is present in `ons-tabbar` I have decided to leave this for after the changes from the react branch have been merged in order to avoid unnecessary conflicts with that branch.


argelius:
Thanks @IliaSky !

PR #1292, opened 2016-03-14
Scroller scrollbar update

IliaSky:
This update removes `ons-scroller` and adds a `.scroller` class with the same css for legacy users.
For the `ons-scrollbar` - it adds a `native` attribute which uses the native scrollbar instead of the js version.
It also removes the infinite scroll from the scrollbar since it was moved to `ons-page`.
I guess we could also remove `ons-scrollbar` completely, if we think that it's unnecessary.


argelius:
@IliaSky I agree that we should remove `<ons-scrollbar>` since it we will never be able to make it work nicely on iOS and the native scrollbar should be enough.

PR #1293, opened 2016-03-14
feat(ons): Select platform from the URL.

frankdiox:
@argelius Should we add this for debugging and showing examples?


argelius:
I think this is good :)

PR #1295, opened 2016-03-15
Splitter reject promises

frankdiox:
@argelius I'm currently just resolving to `undefined` when the menu doesn't open for any reason and resolving to the element itself if it opens. Do you think this is enough or we should change it?


argelius:
I think this is enough! Thanks :+1:

PR #1296, opened 2016-03-16
fix(tests): make tests more stable and update selinium

philolo1:
@argelius when updating node modules on my react_binding, the selinium tests are failing. Sometimes they are failing, because often we try to click an element that is not there yet. This commit fixes a lot of these issues and also updates the drivers for mac and linux.


argelius:
Thanks for the fixes! :+1:

PR #1297, opened 2016-03-16
feat(ons-tab): Modifications to tabbar to make react-binding easier

philolo1:
@argelius I am currently sepeating the code for react that we did, this is the code that you did about the tabbar integration, lets have a look seperatly on what we change here and whether it makes sense.

### DIFF
--- a/core/src/elements/ons-tabbar/index.js
+++ b/core/src/elements/ons-tabbar/index.js
@@ -208,7 +208,17 @@ class TabbarElement extends BaseElement {
       this._compile();
     }
 
-    this._contentElement = util.findChild(this, '.tab-bar__content');
+    for (var i = 0; i < this.firstChild.children.length; i++) {
+      this.firstChild.children[i].style.display = 'none';
+    }
+
+    var activeIndex = this.getAttribute('activeIndex');
+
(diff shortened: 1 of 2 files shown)


philolo1:
@argelius I fixed the tests, should be fine now, question is whether we want to solve this activeIndex problem differently :+1, namly because this activeIndex property will only work if the pages are not loaded.

core/src/elements/ons-tabbar/index.js
```diff
@@ -208,7 +208,17 @@ class TabbarElement extends BaseElement {
...
-    this._contentElement = util.findChild(this, '.tab-bar__content');
+    for (var i = 0; i < this.firstChild.children.length; i++) {
+      this.firstChild.children[i].style.display = 'none';
+    }
+
+    var activeIndex = this.getAttribute('activeIndex');
+
+    if (activeIndex && this.children[1].children.length > activeIndex) {
+      this.children[1].children[activeIndex].setAttribute('active', 'true');
+    }
```
philolo1: @argelius i mean this code, maybe we can do it differently, because this is kind of hacky and won't work if you load pages.

argelius:
I merged this into `nav_mod` instead.

PR #1298, opened 2016-03-16
feat(ons-navigator): Modifications to make react integration easier

philolo1:
@argelius this is the other seperation what we changed for the navigator to make react work, we need to fix the tests first and then we can actually add the react stuff.


philolo1:
@argelius there seems to be much more work than i expected here, so basically we need to get rid of NavigatorPage property and also fix insert/bringtotop etc. More than 40 navigator tests are not working. I will be focusing on fixing these issues, let me know if you also want to take a look and i will push what i have now.

philolo1:
@argelius just cancel this for now, will do a pull request ones it is finished.

PR #1301, opened 2016-03-18
fix(ons-splitter): Fixes #1300.

### LINKED ISSUE #1300: can't exit from app if page has ons-splitter element
Maybe bug:
file: elements/ons-splitter/index.js
line: 304 ( _onDeviceBackButton )
const left = this._getSideElement('left');
const right = this._getSideElement('right');

If <ons-splitter-side> element is set as "left", than test if "right.isOpen()" return error, because "const right" is null.

I removed these lines, and now I'm able to exit from app.

frankdiox:
@argelius


argelius:
Thanks @frankdiox ! :+1:

PR #1304, opened 2016-03-23
fix(ons-popover): fix buggy ios animation

argelius:
@IliaSky I made some changes to the popover since the animation was broken.
I'm not sure if I broke something so please take a look :)

The problem before was that the `popover__container` was very large and since we ran the transitions on it, the origin of the transforms was not inside the popover.


frankdiox:
@argelius This will be fixed in https://github.com/OnsenUI/OnsenUI/pull/1305

PR #1309, opened 2016-03-24
fix(css-components): fix list item padding

argelius:
@frankdiox This fixes the padding for material items. Can you check this and merge if it's OK?


frankdiox:
@argelius Thanks!

PR #1321, opened 2016-03-28
bugfix(navigator) Fixing resetToPage method to not execute always in the callback "onTransitionEnd"

PMoneda:
When the "resetToPage" method is then executed the "callback" "onTransitionEnd" is changed to reset all pages stacked on the navigator and thereby the unstacking of pages is always running even when other methods are invoked such as pushPage that causes an error state in the application, especially if you are working with "back button"

So I created a local variable, "needToDestroyPages", indicating whether the method for unstacking pages should or should not be performed on the "callback" "onTransitionEnd"

### DIFF
--- a/core/src/elements/ons-navigator/index.js
+++ b/core/src/elements/ons-navigator/index.js
@@ -664,13 +664,14 @@ class NavigatorElement extends BaseElement {
     const onTransitionEnd = options.onTransitionEnd || function() {};
 
     options.onTransitionEnd = () => {
-      while (this._pages.length > 1) {
-        this._pages.shift().destroy();
-      }
-      this._pages[0].updateBackButton();
+      
       onTransitionEnd();
     };
-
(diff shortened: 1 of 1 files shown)


IliaSky:
Thank you for the pull request. 

The problem which you are having is a combination of one strange thing on our side and one strange thing on your side. On our side it turns out that we're modifying the options object which you're giving us. I guess you're using the same options object for all methods. Even though yours is kind of strange it's not actually a wrong thing and we shouldn't have bugs because of that.

So the solution is just that we shouldn't modify the object which you're giving us. What you are doing is patching the side effects of that instead of solving the actual problem. Also you are changing the time of execution for `this._pages[0].updateBackButton();` which shouldn't be the case for a bugfix.

I guess we'll just stop modifying the options object instead. It will probably be fixed in the next beta version.

PR #1332, opened 2016-04-01
refactor(ons-scrollbar): Changed to use 'const' or 'let' instead of 'var'.

anatoo:
Trivial fix.

@argelius Merge this changes.


argelius:
Thank you @anatoo! :+1:

PR #1339, opened 2016-04-05
fix(ons-ready): Inner components might not be ready on page init.

frankdiox:
@argelius @IliaSky We still have `ons.ready` after `init` events in some situations but maybe it's not a big problem.


IliaSky:
The back button spec doesn't really seem connected unless I'm missing something.
Otherwise I guess it's fine if it works. We seem to be doing this in a lot of other places anyway...
Feel free to merge it I guess :)

frankdiox:
This also fixes problems with `ons-page` properties like the new `page.options` and `page.name` when using `ons-navigator`. Otherwise those properties wouldn't be available on page `init` event.

PR #1342, opened 2016-04-06
fix(ons-splitter-side): Default side attribute and swipe-target-width.

frankdiox:
@IliaSky


IliaSky:
Looks good - feel free to merge it once it passes CircleCI :)

PR #1354, opened 2016-04-08
Make ons-back-button Angular2 compatible.

anatoo:
Make ons-back-button Angular2 compatible.
ons-back-button now accept the following html:

```
<ons-back-button>
  <span class="back-button__icon"></span>
  <span class="back-button__label">Label</span>
</ons-back-button>
```


argelius:
Thanks @anatoo ! I tried it and it works fine :+1:

PR #1355, opened 2016-04-08
Remove modernizr.

anatoo:
Removed unneeded modernizr.


argelius:
Thanks! I wasn't sure if this was still used. :+1:

PR #1358, opened 2016-04-08
fix(viewport): Fixed viewport overwritten issue, closes #1357.

### LINKED ISSUE #1357: Viewport overwrite bug.
The viewport gets always overwritten, even in it has been specified in the HTML.

andipavllo:
@argelius Please take a look at this when you have time. Also, should I add this in 1.3.15 changelog? There is another important fix that needs to be released so I guess we can release 1.3.15 next week.

### DIFF
--- a/core/vendor/viewport.js
+++ b/core/vendor/viewport.js
@@ -31,14 +31,16 @@
             return;
         }
 
-        if (this.platform.name == 'ios') {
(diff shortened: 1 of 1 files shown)


argelius:
Thanks @andipavllo ! I will merge this.

`1.3.15` has already been released. If you want to release a new verision, please release `1.3.16`.

PR #1365, opened 2016-04-14
feat(ons.notification): Improved API.

frankdiox:
@argelius I think this is what you said :)


argelius:
Thanks @frankdiox ! :+1:

PR #1369, opened 2016-04-18
Auto refresh

argelius:
@IliaSky I noticed that `auto-refresh` attribute was only implemented for AngularJS bindings. I think it makes sense to have it in the core.

This will trigger `carousel.refresh()` when the number of children changes and the `auto-refresh` attribute is defined.


IliaSky:
ok

PR #1372, opened 2016-04-19
docs(*): update doc categories

masahirotanaka:
@argelius I changed doc categories for some components. Please take a look. Here's the result:

image


argelius:
Thanks @masahirotanaka !

The `<ons-splitter-content>` was in the wrong category so I moved it. Looks good now.

PR #1373, opened 2016-04-20
New categories

argelius:
@masahirotanaka Me and @frankdiox had a discussion about how to do the categories.  We thought there was too many categories so we came up with this.

We added all navigation components into `navigation` category. This includes navigator, tabbar and splitter.

We moved popover to `dialog` category since we feel it makes more sense there. We also created a new category called `visuals` where we put icon and progress bars.

What do you think?


frankdiox:
There is also a `material` category now that groups all the material specific components: fab, speed-dial and ripple.

argelius:
I'm closing this

PR #1375, opened 2016-04-20
feat(ons-list-item): Make the attribute `tappable` a boolean and add a property `tap-background-color`

philolo1:
@argelius take a look whether this works, i tested it.


argelius:
Thanks @philolo1 ! :)

PR #1381, opened 2016-04-21
Added prefer-const rule.

anatoo:
Summary:

* Added prefer-const rule on eslint.
* Updated some dependencies.

@argelius Please merge this changes.


argelius:
Thanks @anatoo! :+1:

PR #1394, opened 2016-05-06
fix(ons-navigator): Throw an error if non ons-page is pushed

philolo1:
@argelius This is related to ons-react, but also relevant in the core. We were catching the error to set isRunning to false, but the user does not see any error. Now the error is thrown and the user should see what he did wrong.


philolo1:
@argelius have you taken a look?

argelius:
@philolo1 Thanks! Looks good :+1:

PR #1396, opened 2016-05-06
fix(css-components): Material list item paddings.

frankdiox:
@IliaSky Do you think this could break anything else? I haven't seen any problem. This fixes some padding issue on MD list items.


IliaSky:
well just changing to `border-box` shouldn't have any side effects or at least there are none which I can think of :)

frankdiox:
@IliaSky :+1:

PR #1397, opened 2016-05-06
fix(ons-page): Toolbar animation works with nested components.

frankdiox:
@argelius This is why sometimes the toolbar was not animated on ios-slide. It happened because there were nested components and it couldn't find the toolbar :sweat_smile: Do you think this fix is fine?


frankdiox:
@argelius Nevermind, this could get the toolbar of a splitter side element or some similar problems. For now we can just manually move the toolbar to the top level I guess.

PR #1403, opened 2016-05-10
fix(ons-range): Fix #1391.

### LINKED ISSUE #1391: <ons-range> glitch in Firefox
The track is not covered by the thumb element in Firefox.
screenshot from 2016-05-03 16-49-01

IliaSky:
To me actually it seemed like the behaviour in other browser sounds kind of unreliable. I made some changes so that it works in firefox too. I made the changes only when there is an `ons-range` element since that is when the `.range__left` is created. So for the pure css version there should be no changes. 

I tested it on chrome, firefox, safari. I don't have IE/Edge. @frankdiox can you check if it's fine and merge it if it is. Sorry for bothering you. (-_-)'

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ v2.0.0-rc.5
 ----
  * css-components: Fix material list item paddings.
  * ons-list: Fix [#1401](https://github.com/OnsenUI/OnsenUI/issues/1401).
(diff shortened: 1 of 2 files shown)


frankdiox:
@IliaSky No problem. This is working well in IE/Edge :)

PR #1407, opened 2016-05-11
feat(ons-splitter): Make ons-splitter angular2 compatible.

anatoo:


argelius:
@anatoo Thanks! I will merge this now

PR #1410, opened 2016-05-11
Make ons-popover angular2 compatible.

anatoo:


argelius:
@anatoo Thank you! Looks great!

PR #1422, opened 2016-05-18
feat(ons-splitter-content): Allows navigation.

frankdiox:
@argelius What do you think about this? I just want opinions, this shouldn't be merged. With this we'd have persistent pages, navigation, less flickers and easier maintenance. :dog:
(Try it without AngularJS since I didn't modify the bindings)


argelius:
@frankdiox Looks like a good idea :)

I wonder if it breaks something. I can test it tomorrow

PR #1428, opened 2016-05-25
Changed to use ES2015 on gulpfile.

anatoo:
@argelius Please merge this changes when you have time.


argelius:
Thanks! Looks great

PR #1431, opened 2016-05-25
fix(ons-carousel): Support for ng-repeat. Closes #1168

### LINKED ISSUE #1168: ons-carousel's initial-index does not work with ng-repeat
1.3.14
Looks like `ons-carousel-item`s are generated with `ng-repeat` after the `initial-index` is checked.

frankdiox:
@argelius Carousel was not working at all with `ng-repeat` because the setup was called before Angular compiled all the `ons-carousel-item`. Same for `initial-index` attribute. This is the most painless fix I could think about.

It's also possible to call `refresh()` but it throws events and do some unnecessary stuff so I preferred to call `_setup` and `_saveLastState`. We need to call `_setupInitialIndex` either way.

I removed some duplicated calls from `createdCallback` since we already do the same in `attachedCallback`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 CHANGELOG
 ====
 
+v2.0.0-rc.7
+ * ons-carousel: Support ng-repeat in Angular1 bindings.
+ * ons-carousel: Fix [#1168](https://github.com/OnsenUI/OnsenUI/issues/1168).
+
 v2.0.0-rc.6
 ----
  * ons-splitter: 'load' methods return promises in Angular1 bindings.
--- a/bindings/angular1/directives/carousel.js
+++ b/bindings/angular1/directives/carousel.js
@@ -152,7 +152,7 @@
         CustomElements.upgrade(element[0]);
         return function(scope, element, attrs) {
           CustomElements.upgrade(element[0]);
-          if (scope.$last){
+          if (scope.$last) {
             element[0].parentElement._setup();
             element[0].parentElement._setupInitialIndex();
             element[0].parentElement._saveLastState();
--- a/bindings/angular1/directives/carousel.js
+++ b/bindings/angular1/directives/carousel.js
@@ -152,6 +152,11 @@
         CustomElements.upgrade(element[0]);
         return function(scope, element, attrs) {
           CustomElements.upgrade(element[0]);
+          if (scope.$last){
+            element[0].parentElement._setup();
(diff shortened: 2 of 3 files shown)


argelius:
@frankdiox Thanks! But I'm wondering what you mean by `ng-repeat` not working. It's working fine for me and I tested it on several devices before releasing.

In what way is it not working?

frankdiox:
@argelius Hmmm, in Chrome 50 at least all the items were displayed with `left: 0%` so only the last one was visible. With this change it updates all the items.

argelius:
@frankdiox Does this happen for the examples in the repo? I can take a look again but I am also using Chrome 50 so I'm wondering what the difference between our environments is..

frankdiox:
@argelius I have modified the first example to be something like this:

```
  <ons-template id="example1.html">
    <ons-toolbar>
      <div class="left"><ons-back-button>Back</ons-back-button></div>
      <div class="center">Fullscreen</div>
    </ons-toolbar>

    <ons-carousel initial-index="2" swipeable overscrollable auto-scroll fullscreen auto-scroll-ratio="0.2" animation-options="{duration: 0.1}" var="carousel">
    <ons-carousel-item ng-repeat="item in ['gray', '#085078', '#373B44', '#D38312']" ng-style="{backgroundColor: item}">
      <div class="item-label">{{item}}</div>
    </ons-carousel-item>
    </ons-carousel>

    <ons-bottom-toolbar>
      <div style="float:right" ng-click="carousel.next()">
        <ons-button modifier="quiet">Next</ons-button>
      </div>
      <div style="float:left" ng-click="carousel.prev()">
        <ons-button modifier="quiet">Prev</ons-button>
      </div>
    </ons-bottom-toolbar>
  </ons-template>

```

This also happens for non fullscreen carousels.

frankdiox:
Actually I tried it in Firefox as well and had the same behavior. Maybe the code is wrong somehow?

argelius:
@frankdiox Thanks! I wll take a look at it

frankdiox:
@argelius This was just reported in the forums as well: https://community.onsen.io/topic/458/onsen2-carousel-seems-to-have-problem-with-ng-repeat

argelius:
@frankdiox Ok! I will merge this.

However, I am a bit concerned by the following condition:

```javascript
if (scope.$last) { ... }
```

It feels like this could also happen for directives other than `ng-repeat` that do something similar.

PR #1432, opened 2016-05-26
Animations refactor

IliaSky:
So since the animators before were relatively long and complex and doing almost the same things, this pr references that issue.

If we want we could still write the same syntax using `animit` etc, but for the simple cases we also have a simpler option now. 

There are several new things:

1. There is a `BaseAnimator` class which has an `animate` method. That method can do many animations (similar to animit.runAll). It takes an object with the elements to be animated as well as a second one containing the animations.

So if we use a class to define the animator we can extend the BaseAnimator and use that method.

The animate method's last argument is something similar to 
```
{
  from: {opacity: 1},
  to: {opacity: 0},
  duration: 0.2, // optional values for all animation-options
  restore: true, // whether to restore the original style of the element after the animation
  callback: ... // if it exists
}
```

If instead of one element we want to style multiple elements we can do that by
```
this.animate({a, b}, {
  a: {from: ..., to: ...},
  b: {from: ..., to: ...}
});
```

2. The AnimatorFactory has been tweaked so that it supports new things like platform-specific animations and simpler ways to define an animator.

Right now it can receive:
- an Animator class
- a function which returns an Animator class
- a string (alias for another animator)
- an object

For the last case the object can contain a `defaults` property for default animation options and should have a property for each of the required methods.
In that case an animator will be created from that object, thus receiving the `animate` method of the BaseAnimator.

And for simpler methods it can have just the last argument for the `animate` function instead of a function. 

So it can be something like

```
const Animator = {
  defaults: {duration: 0.3},
  show: function ({element, callback}) {
    this.animate(element, {
      from: {opacity: 0},
      to: {opacity: 1},
      callback
    });
  }
};
```

Currently the animator methods are always being called with an options object as the only argument. The same argument is used in the constructor for the animator class. This allows us to save the options (with the defaults applied) in the constructor and later use it without the user needing to specify more things.

In the most generic case, when we are animating something like `options.element` we can just write the last argument of animate instead of the whole function.

```
const Animator = {
  show: {
      from: {opacity: 0},
      to: {opacity: 1}
  }
};
```

Since most of our elements seem to have getters for their relevant children we can take advantage of that even in this scenario.

```
const Animator = {
  show: {
    _mask: {from: ..., to: ...},
    _thing: {from: ..., to: ..., restore: true, callback: true}
  }
};
```

Since there may be some custom delays and durations in this syntax we can say `callback: true` to note that the callback should be called after this specific part.

3. And to make the animations themselves simpler there is a set of simple animations which can be used.

For example there is a simple object
```
const fade = {
  in: {from: ..., to: ...},
  out: {from: ..., to: ...}
}
```

And some other utilities such as 
- `acceleration` (`translate3d` by 0) 
- `translate({from: '0, 15px', to: '20px, 20px'})`
- `scale({to: 1.3})`
- `reverse(otherAnimation)` - finds and replaces from and to's
- `union(anim1, anim2)`
etc...

And also since most of these functions return an object with properties `from` and `to`, instead of needing to call `util.extend` the following 2 things are equivalent for the `animate` function.

```
{from: a, to: b, restore: true}
{animation: {from: a, to: b}, restore: true}
```

I guess this is the gist of it.

The other changes are:
- a change to `ons-page` and it's children (they get moved around less)
- some generic changes to private property names
- and of course changing the animations to the new format. 
- and since now we can just use objects as animators in some components there are some functions which help to generate those objects.

Another note: the `ios-slide-animator` currently has 2 versions - the old one, which finds labels as well as some strange hacks trying to find specific children which are not labels, and a new one in which the final result is the same, but it doesn't try to find unnecessary children. (The new one does a fade.out on the whole element, while the old one does fade.out on some children, fade.in on others, and sets a background to transparent to see the elements.) I think the new one is better, imo it's much simpler and should be better, but maybe you prefer the old one.

The other thing which bothers me is the requirement of a BaseClass. Basically it has only 1 meaningful method, so maybe just exporting that would be simpler. The necessity of the class comes from getting the default values. If we remove that then we would need to make them manage the default values by themselves which may not be nice.


anatoo:
I looked this changes but I don't want to merge. 
Honestly I can't understand why this refactored code is better than past code.
I think that it become unnecessarily complicated and lower maintenancebility.
@argelius Do you think about this changes?

IliaSky:
@anatoo sorry to hear that. I added a description to explain the changes, guess I should have done that sooner.

The goal was to make the animators simpler. I guess there are some odd cases where they may look more complicated, but I guess generally they are simpler. I guess I may be wrong on that. I may have gone too far into trying to create them shorter to the point where they became more complex than before. In that case what do you think of the state at the `secondary refactor` commit? (It's before I made all the components into objects).

argelius:
@anatoo 

ありがとうございます。マージしないことにしました。
アニメーション周りの変更なので、@anatooさんに見てもらうほうがいいと思いました。

@IliaSky 
I've been thinking about this and decided not to merge it. At least not in the current form.

There are several reasons for this. One is that this is a rewrite rather than a refactor and I don't want to make such a big change at this point since we are currently in RC.

Big changes like this need to be preceded by some discussion and it's important that we all agree on the changes before we start implementing. Maybe this is a fault on mine since I heard that you were working on changes to the animations but I wasn't fully aware of the scope.

Previously when adding new features and making big changes we always start by writing a spec where everyone can chip in and provide feedback. I think this is very important since it avoids situations like this where someone has spent a lot of time making changes but we are unable to merge them.

frankdiox:
@anatoo Do you think we should maintain the animators as they are right now? Or do you have any idea about how you want them?

Currently I think users out of Onsen UI team cannot make their own animators in an easy way. And even for us, when we want to implement a new animator we just copy-paste 150 lines of code from an old animator and modify 3 or 4 values.

argelius:
Looking at the animators resulting from these changes I agree that they are shorter but maybe some things are a bit overkill.

There are some helper functions like `fade.in`, `translate`, etc. that I think are hiding what's actually going on. I think specifying the actual CSS rules is easier to understand.

I also don't like that the names of the keys are used to get the DOM elements. I find that unintuitive.

anatoo:
@frankdiox I'm OK that we create simple animation utility that make animation easily. But I don't think that current animator architecture has big problem.

IliaSky:
Well actually this whole refactor does not change the way in which the animators work. It only acts as an additional wrapper/extension of the current API. In theory an normal animator like the current ones can be used even after the final changes.

So in that case @anatoo what would you consider to be an acceptable animation utility?

In it's current state the one which this PR provides is:
```
animate(options, elementsMap, animationsMap)
```

I guess if using a map for the elements and animations is unintuitive it can just be changed to an array. And for the small helper functions/objects like `fade.in` etc they can also be removed if you feel that they make things harder to understand.

At that point though I'm not sure whether merging would be simpler rather than just doing this from scratch, since almost a month has passed and there were some changes to the core elements themselves (navigator, page etc), which might make merging now a little hard.

So I guess closing the PR and doing the changes from scratch if we want any might be our best course of action.

PR #1435, opened 2016-05-26
feat(ons-navigator):  'data' object for 'popPage'.

frankdiox:
@argelius What do you think about adding this? Some users already asked for it.
The changes to `_pushPage` are just to avoid possible inconsistencies if `bringPageTop` pushes `data` as well.


argelius:
@frankdiox If I understand this correctly it can be used to pass data to the previous page with `popPage()`?

Maybe we can merge this for the next version.

frankdiox:
@argelius Yes, that's the purpose. There were 2 or 3 users who asked for this in the forums or issues. Since it feels more like completing the `page.data` object feature than adding something new I thought we could include it now. But nevermind, it's also fine for the next version.

argelius:
Thanks @frankdiox

PR #1442, opened 2016-06-07
fix(ons-splitter): setImmediate for layout after creation

philolo1:
@argelius somehow in react the layout did not have an effect, this would fix it, but probably there is a better solution.


argelius:
Thanks @philolo1 !

I did some debugging. It turns out that the reason for the bug was that `<ons-splitter-side>` element has a private property called `_width` but it was not set yet.

However, the attribute `width` is defined so I changed the implementation of the `<ons-splitter-side>` slightly instead since I think it's better than using `setImmediate`.

https://github.com/OnsenUI/OnsenUI/pull/1443/files

PR #1445, opened 2016-06-08
fix(css-components): List divider modifiers for MD.

frankdiox:
@argelius A user reported that `nodivider` does not work for Material Design in a private chat. The changes we made for list-item paddings in rc.5 are overwriting `background-image none` of `nodivider`. I just moved the common modifiers to the bottom of the file, I guess this should not break anything.


argelius:
Thanks @frankdiox !

PR #1454, opened 2016-06-14
fix(ons-navigator): Closes #1430

### LINKED ISSUE #1430: Broken page transitions with ons-navigator on ios
__Environment__
OS X El Capitan 10.11.5
Onsen 2 beta 6
iOS (9.3.1 and 9.3.2 on real iPhone 5) iOS 9.3 on 6s emulator 

__Encountered problem__
Page transition animation is broken, tested on (fade and slide) it will briefly flash the next page and then start the animation. Back animation looks fine, but there is no way to know if it has the same “flash” problem - as it would be unnoticeable. 

__Pen__
But there is no problem running from mobile browser
https://codepen.io/anon/pen/WxeRGL?editors=101

__Forum thread I made__
https://community.onsen.io/topic/453/broken-page-transitions-on-ios/2

__How to reproduce__
With latest version and example from https://onsen.io/v2/docs/js/ons-navigator.html

frankdiox:
@argelius This fixes the flickering in Microsoft Edge and should also fix #1430 (at least in Safari).


argelius:
Thanks @frankdiox !

PR #1459, opened 2016-06-15
remove leading / in import statement

matoilic:
I think this leading / is there by mistake, since the ons folder should already be in the root path.


argelius:
Thanks @matoilic ! I will merge this.

PR #1460, opened 2016-06-16
Vue bindings2

elvisqi:


argelius:
@elvisqi Wow, this is amazing! :)

We have actually been thinking about adding Vue.js support so maybe we can cooperate.

PR #1462, opened 2016-06-16
chore(d.ts): Definitely typed.

frankdiox:
@argelius I think this should be enough.


argelius:
Thanks @frankdiox ! :+1:

PR #1465, opened 2016-06-18
Listen for 'input' event instead of 'change' event so we value is updated on every change to the input.

iamjay:
Variable bound to ons-input value doesn't always have the latest value. For example,

````
@Component({
  selector: 'app',
  directives: [ONS_DIRECTIVES],
  template: `
  <ons-page>
    <ons-toolbar>
      <div class="center">Input</div>
    </ons-toolbar>
    <div class="page__background"></div>
    <div class="page__content">
      <div style="padding: 10px">
        <div><ons-input placeholder="Type here" [(value)]="target"></ons-input></div>

        <p>Text: {{target}}</p>
        <ons-button (click)="onClick()">click me</ons-button>
      </div>
    </div>
  </ons-page>
  
})
export class AppComponent{
  target: string = '';

  onClick() {
    console.log(this.target);
  }
}

bootstrap(AppComponent);
```

Typing in the input box then click on the button will result in the old value of input box on the console.

### DIFF
--- a/bindings/angular2/src/directives/ons-input.ts
+++ b/bindings/angular2/src/directives/ons-input.ts
@@ -26,7 +26,7 @@ export class OnsInput implements OnChanges, OnDestroy {
     this._boundOnChange = this._onChange.bind(this);
     this._element = _elementRef.nativeElement;
 
-    this._element.addEventListener('change', this._boundOnChange);
+    this._element.addEventListener('input', this._boundOnChange);
   }
 
   _onChange(event) {


argelius:
Thank you @iamjay !

We will take a look at this PR. Probably we also need to add a case for checkbox and radio buttons since they don' emit the input event.

argelius:
@iamjay I will merge this now since it solves the problem. We will take a look at the radio buttons and checkboxes later.

PR #1468, opened 2016-06-20
fix(ons-switch): Closes #1464.

### LINKED ISSUE #1464: A Switch component inside a Splitter component makes drag events overlap each other.
__Environment__
Onsen 2 beta 6
Javascript
Every platform

__Encountered problem__

A Switch component inside a Splitter component makes drag events overlap each other animating both at the same time.

__How to reproduce__
[Codepen example](https://codepen.io/anon/pen/VjKwdB)

frankdiox:
@argelius `stopPropagation` only works for `onDragStart`. `onDrag` and `onDragEnd` are ignored with `this._ignoreDrag`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ CHANGELOG
 v2.0.0-rc.12
 ----
  * core: Added type definitions.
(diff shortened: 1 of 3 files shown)


argelius:
Thanks @frankdiox !

PR #1470, opened 2016-06-20
Rename internal Cancel event to `dialog-cancel`

philolo1:
@argelius  There is a problem when catching these events in Chrome, as it emits a syntax error. The solution for now is to rename the events since they are not documented.


argelius:
Thanks @philolo1 ! Looks like we have a test for this event so we will need to change that as well.

philolo1:
@argelius thanks for noticing, i fixed the test and updated the commit message!

argelius:
Thanks @philolo1!

PR #1483, opened 2016-06-27
fix(ons.platform): Closes #1482.

### LINKED ISSUE #1482: Automatic styling does not work with cordova-plugin-device in browser
[](Github issues should be used only for bugs and feature requests. For questions please visit https://stackoverflow.com/tags/onsen-ui or https://community.onsen.io. For bugs please use the following template.)

__Environment__
Onsen 2 RC 14
React 15.1.0
React-onsenui 0.5.6

__Encountered problem__
When the `cordova-plugin-device` is loaded, and it's `DeviceProxy.js` is used when testing in the browser, automatic styling does not longer work, but iOS styling is always used.

__Possible cause and solution__
This is because e.g. [the `isAndroid()` function](https://github.com/OnsenUI/OnsenUI/blob/ea1283fdd743637b810ecd98f30cc5beb125939d/core/src/ons/platform.js#L93-101) checks whether the device plugin is loaded (using `typeof device === 'object'`), but in case this turns out to be `browser`, it does not check the `navigator.userAgent` string. 
When one changes the Chrome Device Emulator to a certain device, this is only reflected in the `navigator.userAgent` string, but not in the `device.platform` variable from the cordova device plugin.

Wouldn't it be better if to change the function to something like this, so the `navigator.userAgent` variable is also tested when device.platform is browser:
```javascript
isAndroid() {
    if (this._renderPlatform) {
      return this._renderPlatform === 'android';
    } else if (typeof device === 'object' && !/Browser/i.test(device.platform)) {
      return /Android/i.test(device.platform);
    } else {
      return /Android/i.test(navigator.userAgent);
    }
}
```
This of course should be applied to other `isPlatform()` functions too.

Note: I tested this code only in Chrome@latest, with and without cordova.js/cordova_plugins.js loaded.

frankdiox:
@argelius Please check this when you have some free time.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ CHANGELOG
 v2.0.0-rc.15
 ----
  * ons-navigator: Improved iOS slide animation. Fix [#1457](https://github.com/OnsenUI/OnsenUI/issues/1457).
+ * ons.platform: Fix [#1482](https://github.com/OnsenUI/OnsenUI/issues/1482).
 
 v2.0.0-rc.14
 ----
--- a/core/src/ons/platform.js
+++ b/core/src/ons/platform.js
@@ -75,7 +75,7 @@ class Platform {
   isIOS() {
     if (this._renderPlatform) {
       return this._renderPlatform === 'ios';
-    } else if (typeof device === 'object') {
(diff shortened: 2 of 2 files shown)


argelius:
Looks great!

PR #1487, opened 2016-06-28
Add AngularJS 1.x bindings for `<ons-speed-dial>` and `<ons-fab>`. Closes #1485

### LINKED ISSUE #1485: <ons-speed-dial> not accessible via 'var' attribute
__Environment__
Onsen 2 (v2.0.0-rc.14)
Angular 1.5.0
All platforms

__Encountered problem__
I'm trying to access the `<ons-speed-dial>` element via the `var` attribute, but no matter what I'm doing, I can't seem to reference the element. For reference, I'm trying to invoke `hideItems()` on the element itself. 

All other elements seem to be working, so it appears to be isolated to speed dials. 

__How to reproduce__
Please see this [jsFiddle](https://codepen.io/ruaanvds/pen/pbRXGm?editors=101). 

Is there something I'm missing?

argelius:
@frankdiox Can you take a look at this when you have time?

I've added bindings for `<ons-fab>` and `<ons-speed-dial>`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ v2.0.0-rc.15
  * ons-navigator: Improved iOS slide animation. Fix [#1457](https://github.com/OnsenUI/OnsenUI/issues/1457).
  * ons.platform: Fix [#1482](https://github.com/OnsenUI/OnsenUI/issues/1482).
  * ons-modal: Support showing on init.
+ * ons-speed-dial: Add bindings for AngularJS.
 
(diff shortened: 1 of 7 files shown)


frankdiox:
@argelius Thanks! It loos good. I just noticed a problem with the bindings in general, not only these ones. When controllers are fired, `var` attribute has not created the variable. It works with timeout 0. Any idea?

argelius:
Thanks! Let's investigate that problem. I agree that it's quite annoying

PR #1492, opened 2016-07-02
Added page-loader and added Angular2 directives for ons-tabbar and ons-tab.

anatoo:
@argelius Please review this changes and merge.

Summary of changes: 

 * Added page-loader to customize template loading logic.
 * Added page and pageLoader property on ons-tab.
 * Added angular2 directives for ons-tabbar and ons-tab.

## Added page-loader.

Added ons.PageLoader and ons.defaultPageLoader. This make new abstractions on loading template resources. You can customize template loading with ons.defaultPageLoader. This is useful on writing other framework bindings.

```
// default implementation
ons.defaultPageLoader.internalLoader = function(page, parent, done) {
  internal.getPageHTMLAsync(page).then(html => {
    const element = util.createElement(html.trim());
    parent.appendChild(element);

    done({
      element: element,
      unload: () => element.remove()
    });
  });
};
```

## Added "page" and "pageLoader" property on ons-tab.

Changed ons-tab to have not only "page" attribute but also "page" property. This changes enable to have page location to can't represented as string.

In Angular2, template location is represented as not string but component type.

```
const tab = new OnsTabElement();
tab.page = PageComponent;
```

And changed to have pageLoader property. This enabled to load Angular2 component. 

```
tab.pageLoader = new PageLoader(function(page, parent, done) {
  ... // code for loading Angular2 page components.
});
```

## Added Angular2 directives for ons-tabbar and ons-tab.

Added some directives that enable ons-tabbar to work with Angular2.
This directives enabled to use [page] input on ons-tab.

Example: https://github.com/OnsenUI/OnsenUI/blob/angular2-ons-tabbar/bindings/angular2/examples/tabbar.ts#L55

====

Currently only ons-tab element use this page-loader and have "page" "pageLoader" properties.
If this pull-request is accepted, I will change other elements(e.g. ons-navigator, ons-splitter-side, etc...).


argelius:
Thanks @anatoo !

It looks good. I think we should also use this template loading API for the Angular1 bindings since they are not loading templates correctly from the Angular `$templateCache` service.

PR #1493, opened 2016-07-02
Changed ons-page to accept ".content" and ".background" child elements.

anatoo:


argelius:
Thanks!

PR #1504, opened 2016-07-11
Remove global `ons` object

argelius:
@anatoo This should make it possible to use the Angular 2 bindings without assigning `window.ons` or using a script tag.


argelius:
Looks like webpack is not used to build the package. I will close this and see if I can figure out how to do this correctly.

PR #1507, opened 2016-07-12
fix(ons-navigator): Fixed display conflicts with ons-carousel.

frankdiox:
@argelius This removes all the `style.display` stuff that were not necessary anymore in Onsen 2 navigator since the z-index and element position in the DOM do the job.

`style.visibility = 'hidden'` is needed to fix the flickering in iOS 9.3.


frankdiox:
@argelius @philolo1 Do you think d302145 is enough to fix #1510?
I really think we need to change the API of `_pushPage`. Perhaps we can provide the `update` or `run` function as a parameter in `options` for `pushPage` and fallback to the core one if it's not provided...

And please, let's use Volkswagen for the e2e tests...

frankdiox:
@argelius @philolo1 I give up about e2e. I seriously vote for using [Volkswagen](https://github.com/auchenberg/volkswagen) :doge:

argelius:
@frankdiox I will close this for now :)

PR #1509, opened 2016-07-12
fix(ons-navigator): iOS slide pop animation.

frankdiox:
@argelius Do you know why this was in the animation?
```
-        animit(leavePageDecomposition.toolbar)
-          .queue({
-            css: {
-              background: 'none',
-              backgroundColor: 'rgba(0, 0, 0, 0)',
-              borderColor: 'rgba(0, 0, 0, 0)'
-            },
-            duration: 0
-          }),
```

Is that fixing a bug or anything?
I removed that because when popping pages the toolbar does not fade. That's fine for white toolbars but not when they have different colors.
I adjusted other values after this change and I think now is a bit better.

### DIFF
--- a/core/src/elements/ons-navigator/ios-slide-animator.js
+++ b/core/src/elements/ons-navigator/ios-slide-animator.js
@@ -457,16 +457,14 @@ export default class IOSSlideNavigatorTransitionAnimator extends NavigatorTransi
           .saveStyle()
           .queue({
             css: {
-              transform: 'translate3d(-' + delta.label + 'px, 0, 0)',
-              opacity: 0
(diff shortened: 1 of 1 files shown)


anatoo:
@fran I'm sure that this is not about fixing a bug. This is a thing to animate toolbars that have different background colors.

frankdiox:
@anatoo But that just put the leaving page toolbar transparent instantly, without any fade. With the changes it now fades.

anatoo:
@fran Probably it was broken in someday. If your changes fixed, that would be nice.

argelius:
Thanks @frankdiox !!!

PR #1514, opened 2016-07-20
fix(ons-navigator): 'init' before 'show'. Closes #1513.

### LINKED ISSUE #1513: `show` fires before `init` on 1st page
__Environment__
2.0.0-rc.15
plain js
Browser

__Encountered problem__
`show` event is fired before the `init` event, but only on the first page. All the other pushed pages fire `init` and then `show` as expected. I'm not sure if this is the intented behavior, but it's not consistent for sure.

__How to reproduce__
https://codepen.io/anon/pen/WxraNp - check the console

frankdiox:
@argelius `done` is called asynchronously in `pushPage` when there is animation, but synchronously when animation is `none`. This makes the `show` event be fired before `init`. For example, this also happens in the first page when using `page` attribute but not when the page is directly inside `ons-navigator`.

There are two fixes, call `done` always asynchronously or just fire `show` asynchronously. The only drawback about the first approach is related to this change: 

```
-      nav2 = ons._util.createElement(` <ons-navigator page='backPage'></ons-navigator> `);
+     nav2 = ons._util.createElement(`<ons-navigator></ons-navigator>`);

-      setImmediate(() => {
-        done();
-      });
+     nav2.pushPage('backPage').then(() => done());
```

We are not sure that the page is loaded right after creating the navigator with `page` attribute. Do you think this could be a problem?

### DIFF
--- a/core/src/elements/ons-navigator/index.js
+++ b/core/src/elements/ons-navigator/index.js
@@ -528,18 +528,15 @@ class NavigatorElement extends BaseElement {
    *   [ja]現在表示中のページをを指定したページに置き換えます。[/ja]
    */
   replacePage(page, options = {}) {
-    options = this._prepareOptions(options, page);
-    const callback = options.callback;
+    return this.pushPage(page, options)
+      .then(resolvedValue => {
+        if (this.pages.length > 1) {
+          this.pages[this.pages.length - 2]._destroy();
+        }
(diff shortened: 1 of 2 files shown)


frankdiox:
@argelius After the last changes this keeps 'none' animator synchronous and just fires `show` event asynchronously.

PR #1516, opened 2016-07-20
Page content movement simplification

### LINKED ISSUE #1511: Onsen2 - 100% CPU when 2+ ons-modals are on the page
[](Github issues should be used only for bugs and feature requests. For questions please visit https://stackoverflow.com/tags/onsen-ui or https://community.onsen.io. For bugs please use the following template.)

__Environment__
Onsen 2 rc-15
Angular 1.5.7
OSX - Chrome 51

__Encountered problem__

CPU gets pegged at 100% when an `ons-page` contains 2 or more `ons-modal` components.  

Here is a screenshot of the JS Profile from Chrome Dev Tools.

screen shot 2016-07-15 at 10 30 58 am

When I look at the DOM elements in the Chrome DevTools, I can see the `<div class="page__extra"...>` contents continually changing with new `ons-modal` elements.  Looks like they are stuck in some sort of render loop.

__How to reproduce__
[](If possible provide a working example where the issue is reproduced. For this, go to http://tutorial.onsen.io/ , select a framework and a template and modify or replace it. This will use the latest Onsen UI version released. After you finish, click 'Export to Codepen' button, save it and link it here.)

```html
<ons-template id="item-details.html">
  <ons-page ng-controller="MyController">
    <ons-modal >
        <div>test 1</div>
    </ons-modal>
    <ons-modal >
        <div>test 2</div>
    </ons-modal>    
  </ons-page>
</ons-template>

```
vs No problems with just:

```html
<ons-template id="item-details.html">
  <ons-page ng-controller="MyController">
    <ons-modal >
        <div>test 1</div>
    </ons-modal>   
  </ons-page>
</ons-template>

```

IliaSky:
Now only moves elements a single time only if it is necessary instead of moving them multiple times in both directions until one comes out on top.

Also fixes #1511.

### DIFF
--- a/core/src/elements/ons-page.js
+++ b/core/src/elements/ons-page.js
@@ -364,7 +364,7 @@ class PageElement extends BaseElement {
   }
 
   _elementShouldBeMoved(el) {
-    if (el.className.match(/\bpage__/)) {
+    if (el.classList.contains('page__background')) {
       return false;
     }
     const tagName = el.tagName.toLowerCase();
--- a/core/src/elements/ons-page.js
+++ b/core/src/elements/ons-page.js
@@ -308,25 +308,6 @@ class PageElement extends BaseElement {
     return util.findChild(this, 'ons-toolbar') || nullToolbarElement;
   }
 
-  /**
-   * Register toolbar element to this page.
-   *
-   * @param {HTMLElement} element
-   */
-  _registerToolbar(element) {
-    this.insertBefore(element, this.children[0]);
-  }
-
-  /**
-   * Register toolbar element to this page.
-   *
-   * @param {HTMLElement} element
-   */
-  _registerBottomToolbar(element) {
-    this.classList.add('page-with-bottom-toolbar');
-    this.appendChild(element);
-  }
-
   attributeChangedCallback(name, last, current) {
     if (name === 'modifier') {
       return ModifierUtil.onModifierChanged(last, current, this, scheme);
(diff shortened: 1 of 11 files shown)


core/src/elements/ons-page.js
```diff
@@ -380,14 +363,16 @@ class PageElement extends BaseElement {
     this.setAttribute('_compiled', '');
   }
 
-  _registerExtraElement(element) {
-    let extra = util.findChild(this, '.page__extra');
-    if (!extra) {
-      extra = util.create('.page__extra', {zIndex: 10001});
-      this.appendChild(extra);
+  _elementShouldBeMoved(el) {
+    if (el.className.match(/\bpage__/)) {
```
argelius: Maybe you can use `/\bpage__content/.test(...)` instead since we are not using the match for anything.

core/src/elements/ons-page.js
```diff
@@ -380,14 +363,16 @@ class PageElement extends BaseElement {
...
+      return false;
     }
-
-    extra.appendChild(element);
+    const tagName = el.tagName.toLowerCase();
```
argelius: This is not necessary anymore. `tagName` is always uppercase on modern browsers.

core/src/elements/ons-page.js
```diff
@@ -380,14 +363,16 @@ class PageElement extends BaseElement {
...
+    if (tagName === 'ons-fab') {
+      return !el.hasAttribute('position');
```
argelius: I'm a bit concerned how this would work in Angular 2 and React. Maybe those frameworks will add the attribute later...

We need to test this.

core/src/elements/ons-page.js
```diff
@@ -362,12 +343,14 @@ class PageElement extends BaseElement {
...
 
       util.arrayFrom(this.childNodes).forEach(node => {
-        if (!node.classList || !node.classList.contains('page__background')) {
+        if (node.nodeType !== 1 || this._elementShouldBeMoved(node)) {
           content.appendChild(node);
         }
       });
 
-      this.appendChild(content);
+      const prevNode = util.findChild(this, '.page__background') || util.findChild(this, 'ons-toolbar');
```
argelius: Is this to ensure the order?

IliaSky: yes

argelius:
Seems to work fine, so I will merge it.

PR #1525, opened 2016-07-28
Changed ons-splitter-* to use page-loader.

anatoo:
@argelius Please merge this changes when you have a time.

Summary:
 * ons-splitter-side, ons-splitter-content: Added "page" and "pageLoader" property. Changed to use page-loader instead of `ons._internal.getPageHTMLAsync()`
 *  gulpfile: Fixed that "core" task is stopped when babel fail to transpile once in "serve" task.
 * angular2-binding: Reimplemented the directive for ons-splitter-content and ons-splitter-side.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ CHANGELOG
 
 v2.0.0-rc.16
 ----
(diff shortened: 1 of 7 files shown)


argelius:
@anatoo Thanks! Looks good so I will merge now.

PR #1532, opened 2016-08-01
feat(ons-pull-hook): change pull hook to not mutate the DOM

argelius:


argelius:
@frankdiox: I will merge this if there are no problems.

PR #1534, opened 2016-08-05
Tabs no page attr

argelius:
@anatoo: I added this to support `<ons-tab>` without `page` attribute.

With this change you can render a tabbar like this:

```
<ons-tabbar>
  <div class="tab-bar__content">
    <ons-page>...</ons-page>
    <ons-page>...</ons-page>
    <ons-page>...</ons-page>
  </div>
  <div class="tab-bar">
    <ons-tab ...></ons-tab>
    <ons-tab ...></ons-tab>
    <ons-tab ...></ons-tab>
  </div>
</ons-tabbar>
```

The tabs and pages are already rendered so no need for the `page` attribute.

Is this change ok?


anatoo:
@argelius Supporting `<ons-tab>` without `page` attribute is OK to me. 問題無いと思います

argelius:
Thanks! :+1:

PR #1540, opened 2016-08-10
fix(ons-splitter): Use a different pageLoader. Closes #1537.

### LINKED ISSUE #1537: pageLoader issue
@anatoo @argelius There is a problem with the new `pageLoader` in rc.16. Some user reported that a page loaded in `ons-splitter` fires `init` event twice. Here's why:

```javascript
  load(page, options = {}) {
    ...
    return new Promise(resolve => {

      this._pageLoader.load({page, parent: this}, ({element, unload}) => {

        rewritables.link(this, element, options, fragment => {

          // Remove old content
          this._hide();
          while (this.firstChild) {
            this.firstChild.remove();
          }

          // Attach new content
          this.appendChild(fragment);
          this._show();

          resolve(this.firstChild);
        });
      });
    });
  }
```

Before the last changes, `getPageHTMLAsync` was just creating the page element. After that this method removed the current loaded page in the splitter-content and then it attached the new created element.

However, right now pageLoader creates and attaches the new page element, i.e. at some point there are 2 pages (the old and the new) attached to the splitter. After that it removes everything and reattaches the new element.
If we simply remove the old page before calling pageLoader there will be a flickering since pageLoader is async method and for some miliseconds the splitter-content will be empty.

### Concusion
I think creating the page element and attaching it to the DOM should be separated in two different steps. This would also improve extensibility since a framework could overwrite only 1 step or the 2 of them, and also do whatever it needs between those steps.
Maybe something like `pageLoader.create` and `pageLoader.attach` instead of `pageLoader.load`? I guess Angular could overwrite `create` to use `templateCache` and things like that.

I know this issue in the splitter can be fixed just with a custom pageLoader (instead of `instantPageLoader`), but I want to know what you guys think about all of this. Opinions?

frankdiox:
@anatoo @argelius Possible fix for #1537 . This fires the 4 lifecycle events in the right order and doesn't repeat anything. Do you think adding `replaceContentPageLoader` is fine?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 CHANGELOG
 ====
 
+v2.0.0-rc.18
+----
+ * ons-splitter: Fix [#1537](https://github.com/OnsenUI/OnsenUI/issues/1537).
+
 v2.0.0-rc.17
 ----
  * core: Update Typescript definitions.
--- a/core/src/elements/ons-splitter-content.js
+++ b/core/src/elements/ons-splitter-content.js
@@ -19,7 +19,7 @@ import util from 'ons/util';
 import internal from 'ons/internal';
 import ModifierUtil from 'ons/internal/modifier-util';
 import BaseElement from 'ons/base-element';
-import {PageLoader, replaceContentPageLoader} from 'ons/page-loader';
+import {PageLoader, defaultPageLoader} from 'ons/page-loader';
 import contentReady from 'ons/content-ready';
 
 const rewritables = {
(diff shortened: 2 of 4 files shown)


anatoo:
@frankdiox I think replacing should be in defaultPageLoader. How about the following code?

```
// Default implementation for global PageLoader.
function loadPage({page, parent, params = {}, replace = false}, done) {
  internal.getPageHTMLAsync(page).then(html => {

    if (replace) {
      util.propagateAction(parent, '_destroy');
      parent.innerHTML = '';
    } 
    const element = util.createElement(html.trim());
    parent.appendChild(element);

    done({
      element: element,
      unload: () => element.remove()
    });
  });
}
```

frankdiox:
@anatoo Yes, I think it's better to modify the default ones instead of adding a new one. I just changed it.

argelius:
Thanks @frankdiox !

PR #1542, opened 2016-08-12
Angular1 template cache

argelius:
@frankdiox @anatoo 

This adds support to the `$templateCache` service in AngularJS. What do you think about this solution?

Another solution that would require some more code is that we add something like

```
ons.registerTemplateCache((page) => {
  return Promise.resolve($templateCache.get(page));
});
```

so we don't have to override the function manually.


anatoo:
It's OK to me.

argelius:
@anatoo Thanks! I will merge now

PR #1543, opened 2016-08-12
refactor(ons-navigator): remove arguments from `update` function.

argelius:
@frankdiox This removes the arguments to the `update` function that is passed to `_pushPage` and `_popPage` so the logic should be implemented in the bindings instead.


frankdiox:
@argelius Great! I guess we still need to unify `pushPage` and `_pushPage` or pass the `update` function to `pushPage` in `options`, right?

argelius:
@frankdiox Yes, I think so.

I guess we should add the `update` method as a parameter to the `options` object.

PR #1556, opened 2016-08-19
Added other components API.

anatoo:

Summary:
 * Added Angular2 API for `ons-popover`, `ons-modal`, `ons-alert-dialog` and `ons-dialog`.
 * Renamed `PageParams` to `Params`.
 * Removed unneeded some directives.


bindings/angular2/examples/alert-dialog.ts
```diff
@@ -3,73 +3,79 @@ import {
...
+  ONS_DIRECTIVES,
+  Params,
+  OnInit,
+  OnDestroy
 } from '../src/angular2-onsenui';
 
 @Component({
   selector: 'div',
-  directives: [OnsAlertDialog],
+  directives: [ONS_DIRECTIVES],
```
argelius: I think this parameter has been deprecated in favor of `NgModule`, maybe we should update the examples to work with the latest RC version.

bindings/angular2/examples/modal.ts
```diff
@@ -1,30 +1,75 @@
...
-export class AppComponent {
+export class AppComponent implements OnInit, OnDestroy {
+  private _modal: any;
+  private _destroyModal: Function;
+
+  constructor(private _modalFactory: ModalFactory) {
+  }
+
+  ngOnInit() {
+    this._modalFactory
```
argelius: I like this API :)

:+1:

bindings/angular2/examples/alert-dialog.ts
anatoo: Yes, it should be NgModule. I will update NgModule related code on other pull-request.

argelius:
Looks great. I like the services for creating modals and dialog! :+1: 

I will merge this and as I mentioned in the comments the `directives` parameter of the `@Component` decorator has been deprecated and will be removed in the next RC version.

I think we should update so we are compatible with RC5+ and ignore previous versions since they are not compatible with eachother.

Issue #1524, opened 2016-07-27
Expose animators


argelius:
@OnsenUI/members It should be possible to import animators so they can be extended more easily.

What do you think is a good API for importing the animators?

frankdiox:
Right now creating a new animator is done [like this](https://codepen.io/frankdiox/pen/kXJvQL?&editors=101) (in ES5). I think we should also have a `extend` method to make this easier.

For the core API, perhaps we can move the element classes to `ons` object instead of `window`:
* `ons.OnsNavigatorElement.registerAnimator`
* `ons.OnsNavigatorElement.animators.iOSSlide`
* `var customAnimator = ons.OnsNavigatorElement.animators.iOSSlide.extend({ constructor: ..., push: ..., pop: ...})`

For ES6 imports:
* `import { OnsNavigatorElement } from 'onsenui';`
* `OnsNavigatorElemet.animators.iOSslide`
* `OnsNavigatorElemet.registerElement`

Maybe we can drop `Ons` prefix.

The common use case is to change 1 or maybe 2 animations, so I guess it's not necessary to put all the animators togethers in only one object, although it's also a possibility.

philolo1:
Should animators depend on the object? I think it would make more sense to
make them more general and expose them in an ons object if its possible.

IliaSky:
@philolo1
Basically they don't really depend much on the object - the question is more like where would they feel natural to be accessed from. 

The splitter, tabbar and navigator animations are special since they operate on different objects.

In theory if we put them in the same place I think we may get users confused a little.

You are correct that the animations for `ons-modal`, `ons-popover`, `ons-dialog` and `ons-alert-dialog` should be interchangeable since they have the same methods and arguments.

Maybe a similar option to what you suggest would be to put them in `ons.animators`, but still group them there like
`ons.animators.navigator` `ons.animators.general` etc,
but I'm not sure whether this makes more sense than just `ons.OnsNavigatorElement.animators`.

@frankdiox 
For the "ons object instead of window" +1 I also think that.

I would be fine with dropping either Ons or Element from `ons.OnsNavigatorElement`, but maybe it seems more explicit and readable for some people.
I think it would be best if we could do
```import {Navigator} from 'onsenui.elements'```
but I don't think that is possible. Anyway I think both any change and no changes in this regard would be fine.

[...]

PR #1559, opened 2016-08-24
feat(ons-navigator): Provides extendable animators.

frankdiox:
@argelius @anatoo As specified in #1524, this expose animators as `window.OnsNavigatorElement.animators` and allow `animator.extend({ ... })`. I think this should be enough since any property of the animator can be modified (`push`, `pop`, `_decompose`, `backgroundMask`, etc.).

`OnsNavigatorElement` is not exported yet, that can be addressed in a different PR.

If you want you can manually test this with:

```
var custom = window.OnsNavigatorElement.animators["fade-ios"].extend({
   push: function(a, b, c) {
     console.log('asereje ', this.duration);
     c();
   }
});

window.OnsNavigatorElement.registerAnimator('custom', custom);
$('ons-navigator').pushPage('page1.html', { animation: 'custom' });
```

Tell me if you see any problem with this or find anything to improve.


anatoo:
It looks no problems to me.

argelius:
Looks great! :+1:

PR #1567, opened 2016-08-29
Custom elements v1

argelius:
This upgrades to Custom Elements v1. I used [this polyfill](https://github.com/WebReflection/document-register-element). It seems to work better than the Polymer polyfill.

@anatoo @frankdiox Please take a look when you have time. The changes are very large but mostly simple refactoring.

@asial-matagawa Can you please verify that there are no issues in browsers other than Chrome?


core/src/elements/ons-splitter-side.js
```diff
@@ -487,12 +491,16 @@ class SplitterSideElement extends BaseElement {
...
     this._gestureDetector.dispose();
     this._gestureDetector = null;
   }
 
+  static get observedAttributes() {
+    return this._watchedAttributes;
+  }
+
   attributeChangedCallback(name, last, current) {
     if (this._watchedAttributes.indexOf(name) !== -1) {
```
frankdiox: I guess this kind of thing is not necessary anymore, right?

core/src/elements/ons-back-button.js
```diff
@@ -74,15 +74,17 @@ class BackButtonElement extends BaseElement {
...
-  createdCallback() {
-    contentReady(this, () => {
-      if (!this.hasAttribute('_compiled')) {
-        this._compile();
-      }
+  constructor(self) {
+    self = super(self);
+
+    contentReady(self, () => {
+      self._compile();
```
anatoo: Why you are using "self" variable? According to custom elements spec(https://www.w3.org/TR/custom-elements/), you should able to use "this" variable in constructor.

argelius: It didn't work when I was creating a normal constructor. Probably because of this:

https://github.com/WebReflection/document-register-element#upgrading-the-constructor-context

core/src/elements/ons-splitter-side.js
argelius: @frankdiox Thanks! Good catch. I will remove the line.

core/src/elements/ons-back-button.js
anatoo: "self" variable is ugly... 😞  Okay.

argelius: We can do this instead: 
https://github.com/WebReflection/document-register-element#skipping-the-caveat-through-extends

So we use `init` method instead of `constructor` but we can use `this`.

Which do you prefer? I actually prefer the using `init` method since we need less code and can use `this` variable.

frankdiox: In that case we are forced to always extend the same element, right? We cannot do `extend HTMLLabelElement`, for example. Not sure if this is well supported in the polyfill anyway.

argelius: @frankdiox Yes, I don't think that is supported. We already extend `BaseElement` so I think we can add the ugly constructor there and then use `init` in the other components.

anatoo: I prefer `init` way. Will you refactor?

argelius: Yes, I will refactor it. Thanks for the feedback! :+1:

argelius: @anatoo I have refactored it now.

argelius:
I will merge this now.

PR #1570, opened 2016-08-31
Changed to provide umd version on angular2-binding.

anatoo:
@argelius @frankdiox Could you review this changes?

Summary:
 * Changed to provide umd version on `bindings/angular2/dist/bundles/angular2-onsenui.umd.js`.
 * Added  `codepen-template` directory to run "angular2-onsenui.umd.js" on codepen or tutorials.
 * Added `gulp serve-codepen-template` task to run codepen template.


frankdiox:
@anatoo Great! Just one thing, instead of `codepen-template` maybe it can be called `umd-template` or something like that? I mean it's not really related to Codepen, right?

anatoo:
@frankdiox Thanks for the feedback. I just renamed.

argelius:
@anatoo Looks good!

PR #1572, opened 2016-09-06
Make CSS Framework RTL

hadjloo:
only css-components got RTL version, the rest is yet to complete

I could't understand how to make the rest of the framework RTL, it is really messy, you have to clean your code and do some refactor

please let me know whcih other files (js - css) should be changed to have a full and complete RTL version of the framework, you can replace main-rtl.css with your current main.css, and everything should be work fine in Right to left mode

don't forget that I didn't changed any value, I just used RTLit mixins to make your LESS files RTL

Final designer does not need to change any thing too, he has to design just like current LTR mode, when he uses the main-rtl.css, the magic will happen and everything should work in RTL mode

Not to mention that RTL is not just about direction or text-align, it is far more than these, I would be happy to make the entire framework RTL, but I really need to know how it is working,

let me know if you had any questions

You can check my github profile and see other css frameworks that I worked on their RTL form

Regards


argelius:
Hi @hadjloo !

It looks like you converted the less file for the website. The Onsen UI components are written in Stylus and you can find them here: https://github.com/OnsenUI/OnsenUI/tree/master/css-components/components-src/stylus/components

Does the mixin also support Stylus?

hadjloo:
I've changed all .styl files in component-src, and added RTLit
stylus-lang mixins in this folder

Also, Added a new index-rtl.styl to provide a full feature RTL index,

You have to consider that I've added 3 variables in index.styl and
index-rtl.styl, these 3, are RTLit variables and are using for RTLit
mixins

Let me know if I should change Other files too

Regards

hadjloo:
@argelius did you check the new pull request? I convertd all files in the mentioned directory and I'm looking forward to your answer

let me know if there is any problem

argelius:
@hadjloo Thanks for the PR. I will take a look when I have time.

Btw, does this take flexbox into account? We use flex-index in some places to force the order of the layout.

ajck:
@hadjloo @argelius What happened to this PR please? Does it work for RTL? If not, what would be needed to get it to work? Or does RTL work in OnsenUI anyway now? :)
Many thanks.

emccorson:
@ajck Onsen UI does not currently support RTL. It looks like this PR is outdated. I'm maintaining Onsen UI at the moment, although I don't have any experience of RTL. We would welcome a PR to support this and I can provide assistance if needed. Otneriwse, it is still on the list but there are other issues that take precedence at the moment.

hadjloo:
@ajck @emccorson the RTL version belongs to version 2 and since version 3 was not backward compatible, then I decided to not spend more time on this project. I would also highly recommend you take a look at original FLUENT components by Microsoft [](https://developer.microsoft.com/en-us/fluentui)

PR #1576, opened 2016-09-08
Fix webpack error

argelius:
@anatoo @frankdiox 

The minified Promise polyfill was causing a warning when people imported Onsen UI with webpack. This solves the problem.


anatoo:
いいんじゃないでしょうか

PR #1577, opened 2016-09-08
Added doc comments on angular2-binding.

anatoo:
@argelius I added Japanese doc comments and added `@example` tags on angular2-binding. Could you fill English empty doc comments?


argelius:
Thanks @anatoo! I have added translations so I will merge now.

PR #1582, opened 2016-09-14
Innerhtml helper

argelius:
@frankdiox @anatoo 

Please refer to this discussion: https://github.com/WebReflection/document-register-element/issues/75

On older devices elements are not upgraded when added using `innerHTML`. This helper function will upgrade them instantly:

On Firefox:

```
var a = document.createElement('div');
a.innerHTML = '<ons-switch></ons-switch>'; console.log(a.firstChild.checked); // undefined
innerHTML(a, '<ons-switch></ons-switch>'); console.log(a.firstChild.checked); // false
```

### DIFF
--- a/core/vendor/innerHTML.js
+++ b/core/vendor/innerHTML.js
@@ -1,0 +1,93 @@
+// see https://github.com/WebReflection/document-register-element/issues/21#issuecomment-102020311
+var innerHTML = (function (document) {
+
+  var
+    EXTENDS = 'extends',
+    register = document.registerElement,
+    div = document.createElement('div'),
+    dre = 'document-register-element',
+    innerHTML = register.innerHTML,
+    initialize,
+    registered
+  ;
+
+  // avoid duplicated wrappers
+  if (innerHTML) return innerHTML;
+
+  try {
+
+    // feature detect the problem
+    register.call(
+      document,
+      dre,
+      {prototype: Object.create(
+        HTMLElement.prototype,
+        {createdCallback: {value: Object}}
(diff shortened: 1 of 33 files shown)


core/vendor/innerHTML.js
```diff
@@ -0,0 +1,93 @@
...
+
+  } catch(meh) {}
+
+  // in other cases
+  registered = [];
+  initialize = function (el) {
+    if (
+      'createdCallback' in el         ||
+      'attachedCallback' in el        ||
+      'detachedCallback' in el        ||
```
anatoo: Should you replace `attachedCallback` and `detachedCallback` with `connectedCallback` and `disconnectedCallback`?

argelius: @anatoo The v1 polyfill maps `connectedCallback` -> `attachedCallback` so the helper still works even if it references `attachedCallback`:

image

frankdiox:
@argelius Great! Do we support Android 4+ again after this? 😅 
Just one concern... Should we now expose some kind of `ons.createElement('<ons-> ... </ons->')` that uses this polyfill internally?

core/vendor/innerHTML.js
anatoo: Oh, I see.  OK.

argelius:
@frankdiox 

`ons._util.createElement` uses this helper function now... I think we should just expose it as `ons.createElement`.

frankdiox:
@argelius  Yes, I think that's fine.

Issue #1580, opened 2016-09-12
ons-back-button doesn't work after pushpage with splitter in ios cordova


juspriet:
__Environment__
Onsen 2.0.0-rc.18
jquery
iOS 9.3.5

__Encountered problem__
I have a ios cordova application with splitter and ons-navigator inside ons-splitter-content.
After pushPage, ons-back-button doesn't work.
And pushPage animation doesn't slide !

argelius:
@juspriet Can you create a simple example that reproduces this bug?

juspriet:
here is a working exemple.
No need to put a ons-splitter component.
The problem seems to be when there is no ons-toolbar on the first page but there is one in the second page....


[Archive.zip](https://github.com/OnsenUI/OnsenUI/files/467226/Archive.zip)

munsterlander:
@juspriet I will look at your code, but below is fully functioning code on both platforms for me.

```
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="utf-8" />
    <meta name="apple-mobile-web-app-capable" content="yes" />
    <meta name="mobile-web-app-capable" content="yes" />
    <meta http-equiv="Content-Security-Policy" content="default-src * data:; style-src * 'unsafe-inline'; script-src * 'unsafe-inline' 'unsafe-eval'">
    <link rel="shortcut icon" href="favicon.ico" type="image/x-icon" />
    <link href='https://fonts.googleapis.com/css?family=Roboto:400,300italic,300,400italic,500,700,700italic,500italic' rel='stylesheet' type='text/css'>
    <title>Onsen UI Forum Help by Munsterlander</title>

    <link rel="stylesheet" href="https://cdn.rawgit.com/OnsenUI/OnsenUI-dist/2.0.0-rc.18/css/onsenui.css" type="text/css" media="all" />
    <link rel="stylesheet" href="https://cdn.rawgit.com/OnsenUI/OnsenUI-dist/2.0.0-rc.18/css/onsen-css-components.css">
    <script src="https://cdn.rawgit.com/OnsenUI/OnsenUI-dist/2.0.0-rc.18/js/onsenui.js"></script>
    <script src="components/loader.js"></script> 

  </head>
  <body>
<ons-navigator page="home.html" id="myNav"></ons-navigator>
<ons-template id="home.html">
    <ons-page>
        <ons-button onclick="document.getElementById('myNav').pushPage('page1.html');">Push Page</ons-button>
    </ons-page>
[...]

PR #1583, opened 2016-09-14
Fixed dependency problem between ons-page and ons-toolbar.

asial-matagawa:
@argelius @frankdiox @anatoo

This patch fixes at least #1580.
`ons-page.js` calls `document.createElement('ons-toolbar')` ahead of `customElements.define('ons-toolbar', ToolbarElement)` (in `ons-toolbar.js`), so I added the following statement:

```
import 'elements/ons-toolbar';
```

If there is no problem, could you merge this patch?


argelius:
Thanks @asial-matagawa ! Great solution. I will merge this now.

PR #1589, opened 2016-09-16
fix(css-components): Toolbar modifiers.

frankdiox:
@argelius toolbar's transparent modifier was duplicated and `noshadow` is not in the docs. I think this would fix it.


argelius:
Thanks @frankdiox !!

PR #1596, opened 2016-09-20
fix(ons.notification): Wait for compilation. Closes #1595

### LINKED ISSUE #1595: ons.notification layout is broken in Android
See attached file.

__Environment__
Onsen 2.0.0

ons-notification

frankdiox:
@argelius @anatoo Commit 75f187ebcf972893364927087e6e5a58ea602b56 broke `ons.notification` styles (see #1595 ). The reason is that after this change onsAlertDialog's `contentReady` will postpone the element compilation until the next cycle, thus making `modifier = modifier || dialogElement.getAttribute('modifier');` to be `null`.

This also means that the rest of the function modifies the element content before it is compiled. Or that is what I thought until I noticed something peculiar. The element is compiled after it is attached to the DOM because in onsAlertDialog's `connectedCallback` we casually use `contentReady` for something different, and at this point it runs the callback queue synchronously (thus compiling the element). This feels like an unknown feature for me 😅 

I think maybe we could use `contentReady` for waiting until the element is compiled. I just modified it slightly to make it more elegant. There are other easy fixes (a bit uglier) but I thought this could be nicer. What do you think?

As a side idea, perhaps we can create `ons.createElement('<ons-element attributes>content</ons-element>')` that internally calls: 

```javascript
let el = document.createElement('ons-element')
el.setAttribute( -attributes- );
innerHTML(el, content)
contentReady(el);
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ CHANGELOG
 v2.0.1
 ----
  * angular1: Fix [#1588](https://github.com/OnsenUI/OnsenUI/issues/1588).
+ * ons.notification: Fix [#1595](https://github.com/OnsenUI/OnsenUI/issues/1595).
 
 v2.0.0
 ----
--- a/core/src/ons/content-ready.js
+++ b/core/src/ons/content-ready.js
@@ -41,7 +41,7 @@ function consumeQueue(element) {
   callbacks.forEach(callback => callback());
 }
 
-export default function contentReady(element, fn) {
+export default function contentReady(element, fn = () => {}) {
   addCallback(element, fn);
 
   if (isContentReady(element)) {
--- a/core/src/ons/notification.js
+++ b/core/src/ons/notification.js
@@ -16,6 +16,7 @@ limitations under the License.
 */
 
 import util from './util';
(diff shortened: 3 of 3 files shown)


anatoo:
@fran It looks good to me. Thanks!


>As a side idea, perhaps we can create ons.createElement('<ons-element attributes>content</ons-element>') that internally calls:
```
let el = document.createElement('ons-element')
el.setAttribute( -attributes- );
innerHTML(el, content)
contentReady(el);
```

Good idea.  You should implement this.

frankdiox:
I got a working version of this proposed `createElement` but unfortunately there's something I didn't think about. Some components are not using `contentReady` (because they don't need content) so they are compiled directly before I set the attributes. So unless we want to implement `attributesReady` I guess we should forget about this version of `createElement` 😅

Do you have any suggestion for this?

Another secondary problem is that it's not possible to correctly parse HTML/XLM with RegExp and detect errors.

anatoo:
Oh that's right :sweat: Let's leave it as it is on `createElement`.

argelius:
Thanks @frankdiox ! Sorry for breaking this... :sweat:

PR #1611, opened 2016-10-06
Add codepen tags in Angular 2 binding.

anatoo:
@argelius @masahirotanaka Added codepen tags in Angular 2 binding. Merge this changes if there is no problem.

### DIFF
--- a/bindings/angular2/src/directives/ons-input.ts
+++ b/bindings/angular2/src/directives/ons-input.ts
@@ -15,6 +15,7 @@ import {
  * @element ons-input
  * @directive OnsInput
  * @selector ons-input
+ * @codepen VKzqyX
  * @description
  *   [en]Angular 2 directive for `<ons-input>` component.[/en]
  *   [ja]`<ons-input>`要素のAngular 2ディレクティブです。[/ja]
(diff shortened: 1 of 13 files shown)


argelius:
Thanks @anatoo !

frankdiox:
@anatoo @argelius Aren't we using the interactive tutorial for this instead?

anatoo:
@frankdiox Tanaka-san said to use codepen until interactive tutorials are ready.

frankdiox:
@anatoo Yes but it is ready already since you implemented the tutorials. It's even in the webpage: https://onsen.io/v2/docs/angular2/ons-navigator.html

We only need to specify the tutorial in the componens that has a different URL from the vanilla tutorials. For example, [`ons-alert-dialog`](https://onsen.io/v2/docs/angular2/ons-alert-dialog.html) (it shows `ons-dialog` instead). The rest are automatically converted, we do something like --> `hoge.replace('vanilla', 'angular2');`, so if they are under `Reference` category and has the same ID it works.

anatoo:
@frankdiox Oh, I see. I thought that tutorials is not ready because some tutorials have no tutorial contents. 

@argelius Sorry, would you revert this pull-request changes?

argelius:
@anatoo I have reverted this now.

anatoo:
@argelius arigato!

PR #1612, opened 2016-10-06
Removed "_compiled" attributes from core.

anatoo:
@argelius I removed all `_compiled` attributes from core. Please review this changes.


argelius:
Thanks @anatoo!

PR #1624, opened 2016-10-14
Fix switch

argelius:
@frankdiox This fixes the regression with the switch. Please take a look when you have time.


frankdiox:
@argelius Looks good, thanks!

PR #1625, opened 2016-10-14
feat(ons-tab): added ability to add badge to tab item

co3k:
Added support to append a badge to tab item like UITabBarItem of iOS SDK.


argelius:
@co3k This looks great!

I've pushed the branch and made a new PR:
https://github.com/OnsenUI/OnsenUI/pull/1625

Will merge it soon.

Just one thing: There is no need to update the generated files.

Issue #1629, opened 2016-10-16
onsenui + jquery: different dom behaviour with ons-page and ons-tabbar


bdedardel:
[](Github issues should be used only for bugs and feature requests. For questions please visit https://stackoverflow.com/tags/onsen-ui or https://community.onsen.io. For bugs please use the following template.)

__Environment__
Onsen 2.0.1
Angular 1.5.8
Linux / Firefox

Hi,

I've prepared some examples with some issues I 've encountered.
This one has blocked my current development and I don't know how to bypass it.

I have different behaviour since I use ons-tabbar.

- with ons-page, I could modify dom with jquery, for example:

```
$('#mydiv').html('<div>TEST</div>');
```

- but with ons-tabbar, it doesnot work.

 

__Encountered problem__

__How to reproduce__
I put an example in my github repo:

[bdedardel/cordova-test/tree/master/onsenui/bug1629](https://github.com/bdedardel/cordova-test/tree/master/onsenui/bug1629)

argelius:
Thanks for reporting @bdedardel ! I will take a look at this now.

argelius:
@bdedardel I think this is related to this: https://github.com/OnsenUI/OnsenUI/issues/1584

This has been fixed in this PR: https://github.com/OnsenUI/OnsenUI/pull/1634

Hopefully we can merge and release this pretty soon.

bdedardel:
Resolution confirmed with onsenui 2.0.5.
It was my mistake by imbricating *ons-template* inside *ons-template*!

```
<ons-template id="show-onstabbar.html">
	<ons-page ng-controller="MyModelController as mymodelCtl"
		id="showonstabbar-page">
		<form id="showonstabbar-form" name="showonstabbarForm" class="onstabbar-form"><!-- 
				novalidate="true" -->
			<!-- ===== begin toolbar: show-onstabbar ===== -->
			<ons-toolbar id="show-onstabbar-toolbar" style="background-color:#333;">
				...
			</ons-toolbar><!-- end toolbar: show-onstabbar -->
			<!-- ===== begin tabbar: show-onstabbar ===== -->

<ons-template id="tab-form.html">
	<ons-page id="tab-form-page">
		...
	</ons-page>
</ons-template>

			<ons-tabbar id="show-onstabbar-tabbar" position="top">
				<ons-tab label="FORM" page="tab-form.html" active="true"
					style="background-color:#333;">
				</ons-tab>
[...]

Issue #1584, opened 2016-09-14
ons-tabbar inserts duplicated pages on iOS / mobile Safari


asial-matagawa:
Reported by @iqmeta in #1580.

> Also there is a new **tab-bar problem with iOS / mobile Safari,** on Chrome it's working fine:
> 
> ```
> document.addEventListener('init', function (event)
> {
> 	var page = event.target;   
> 	console.log("init:" + page.id);
> }
> ```
> 
> 
> **Chrome console is correct:**
> ```
> init:tabBar
> init:menu
> init:gruppenPage
> init:home
> ```
> 
> **On iOS gruppePage is added 3x times:**
> ```
> init:menu
> init:gruppenPage
> init:gruppenPage
> init:gruppenPage
> init:home
> ```
> 
> **Markup Tabbar** no where else in code gruppen.html was referenced
> ```
>  <ons-tabbar id="tabs" animation="slide">
>         <ons-tab active page="home.html" id="alarm">
>             <div class="tab-bar__icon">
>                 <ons-icon icon="ion-ios-bell" size="33px" style="line-height:0px"></ons-icon>
>             </div>
>         </ons-tab>
>         <ons-tab id="gruppen" page="gruppen.html">
>             <div class="tab-bar__icon">
>                 <ons-icon icon="ion-radio-waves" size="33px" style="line-height:0px"></ons-icon>
>             </div>
>         </ons-tab>
> </ons-tabbar>
> ```
> **gruppen.html**
> ```
> <ons-page id="gruppenPage">
>     <div id="testX">X</div>
>     <ons-list id="groupList">
>     </ons-list>
> </ons-page>
> ```
> 
> ```
> tabs.addEventListener('postchange', function(event)
> {
> 	$("#textX").text("Y") <= works only on chrome. on iOS not! because tabBar added 3 times the gruppenPage.html. Th
> }
> 
> ```
> 
> **Screenshot from Cordova Remote Debugger**
> ![image](https://cloud.githubusercontent.com/assets/7533842/18483683/67da9054-79e5-11e6-86d9-a00c98e6dcf8.png)
> 
> 
> 
> Going back to R17 for now. There this works flawless ;-) Hope this helps a bit. Great work on onsen by the way :+1: .
> 
> 
> Update:
> With R17 i have double inits with splitter too:
> ![image](https://cloud.githubusercontent.com/assets/7533842/18491964/435e4ad0-7a09-11e6-8133-bd3ee42133d6.png)
> 
[...]

PR #1634, opened 2016-10-17
Tabbar refactor

argelius:
@anatoo @frankdiox 

This removes a lot of complex code from tabbar and makes it easier to understand. This code was handling loading pages and things like that but I think it's better to only support persistent pages.

It also fixes an issue that happened when `connectedCallback` was called more than one time for `<ons-tab>`. I think the following issues can be closed:

https://github.com/OnsenUI/OnsenUI/issues/1629
https://github.com/OnsenUI/OnsenUI/issues/1584

@anatoo What do you thing about the `loadPage` method? It attaches a new DOM element but it is never removed. If you don't mind I would like to remove it or maybe add a deprecation warning.


anatoo:
@argelius It's OK to make deprecated.

Issue #1638, opened 2016-10-17
notification prompt "Capturing backbutton-handler is failure."


Wieltje:
__Environment__
Onsen 2.01
Cordova
Android 6

__Encountered problem__
I'm using ons.notification.prompt with cancelable = true, Pressing the back button leads to "Capturing backbutton-handler is failure." instead of closing the prompt.

argelius:
@Wieltje I've made a fix for this and it will be released with the next version.

In the meantime you can use this CSS to fix it:

```
ons-alert-dialog {
    position: absolute;
    top: 0;
    left: 0;
   bottom: 0;
   right: 0;
   overflow: hidden;
   z-index: 2;
 }
```

Wieltje:
Thanks for your quick response and fix! It seems however that the approach is different than when cancelling by clicking the background, as when using the workaround, selection markers that are applied on the notification.prompt don't disappear when cancelling the prompt with the back button. Is this something you recognize?
![image](https://cloud.githubusercontent.com/assets/22632746/19471274/c3ca7912-9522-11e6-9995-9e2e1f2026fc.png)

argelius:
@Wieltje Thanks for the quick response. I will investigate this as well.

argelius:
@Wieltje Actually this doesn't happen when I tried it. I was using an Android 6.0.1 device, what version are you using?

Wieltje:
I'm also using 6.0.1. 
I've tried to isolate and reproduce the issue and so far it seems that there's no problem when the prompt is called from my ons-navigator mainpage but does when called from another page.
I'll have another look at it later.

Wieltje:
I managed to isolate the problem to [this code](https://codepen.io/anon/pen/NREKmb?).

As long as the page does not overflow, the problem exists. After filling the div so that the page overflows, problem is gone. Same behavior on android 5.01 and 6.01

@argelius Would you mind to try this out to reproduce?

PR #1639, opened 2016-10-18
Fix device back button

argelius:
@anatoo @frankdiox 

This fixes an issue with the device back button and dialogs. The function that finds which element to run the back button handler relies on the z-index if there are more than two components on the same level.

This changes the CSS of `ons-dialog` and `ons-alert-dialog` so they have the same z-index as other components.

Fixes this issue: https://github.com/OnsenUI/OnsenUI/issues/1638


anatoo:
👍

Issue #1636, opened 2016-10-17
ion icons are misaligned in Android toolbar


rancor1223:
__Environment__
Onsen 2.0.1
Android

__Encountered problem__
_ion icons_ are misaligned in the toolbar (and maybe elsewhere?) when using Android.

```
.ons-icon--ion {
    line-height: 0.75em;
    vertical-align: -25%;
}
```

Problem is the `vertical-align: -25%;`, which is supposed to center it vertically. It does so on iOS, but it misaligns it on Android.

__How to reproduce__
I [wasn't able to reproduce this in a CodePen](https://codepen.io/anon/pen/NRLBKa) as it doesn't let me change platform to Android. However, it's possible to reproduce it in the [Onsen Tutorial Environment](http://tutorial.onsen.io), so you can just paste it there and change the platform to Android.

argelius:
@rancor1223 Thanks for reporting!

I will take a look at this issue.

PR #1640, opened 2016-10-18
Fix 1636

argelius:
@anatoo @frankdiox 

Fixes this issue:
https://github.com/OnsenUI/OnsenUI/issues/1636

There was some CSS that fixed the alignment only for Ionicons, however in the case where they are used in the toolbar buttons it is not necessary since they are displayed vertically aligned anyway.


frankdiox:
@argelius Thanks!

PR #1642, opened 2016-10-18
Fix 1620

argelius:
@anatoo @frankdiox 

This fixes a memory leak in the AngularJS bindings when `resetToPage` was used.


anatoo:
Looks good to me 👍

frankdiox:
@argelius Thanks!

PR #1648, opened 2016-10-24
Improve the response of ons-pull-hook.

asial-matagawa:
@anatoo @frankdiox @argelius 
I found a bug where ons-pull-hook does not work when we swipe up/down a screen too fast.
When we swipe up/down too fast, the gesture detector occasionally dispatches `dragleft` or `dragright`.

So it seems that we have to let the pull hook listen to `dragleft` and `dragright` as well as `dragup` and `dragdown`, so I added some lines.
But since it causes conflict with ons-splitter, I put some additional lines to avoid it.

I am not sure this PR causes some problem or not.
Could you review this code?


asial-matagawa:
This bug could be reproduced in Chrome on PC and Safari on iOS 9.3.2.

argelius:
Thanks @asial-matagawa ! Looks great. I will merge this now.

PR #1652, opened 2016-10-26
refactor(ons-switch): Use contentReady.

frankdiox:
@argelius I think this should be enough for Vue2.


frankdiox:
@anatoo Can you check commit a2a12a4 ? I need to do that to make it work with the previous commits in Angular 2. Do you think that can be a problem? It's doing the same as ons-input.

anatoo:
@fran It should be OK.

argelius:
@frankdiox This looks good and it doesn't break anything. I will merge now.

PR #1659, opened 2016-10-27
Remove deprecated gulp plugin `lazd/gulp-karma`.

asial-matagawa:
@anatoo @argelius @frankdiox 

Since [`lazd/gulp-karma`](https://github.com/lazd/gulp-karma) has been deprecated as of 2015/10/02, I have removed `lazd/gulp-karma` from this repo and started using Karma directly according to [the official guide](https://github.com/karma-runner/gulp-karma).
Also, I have added some messages.

If there is no problem, could you merge this PR?

PR #1660, opened 2016-10-27
Add functionality to separately launch Karma server for each spec file.

asial-matagawa:
@anatoo @argelius @frankdiox 

Note: This PR has conflicts with #1659.

Since some browser breaks the unit testing itself, I need functionality to separately launch Karma server for each spec file.
This PR contains a new gulp task named `core-test-separately` for debugging purpose, which launches Karma server for each spec file or arbitrary spec files specified by user.
It also contains a dirty hack to pass variables from `gulpfile.babel.js` to `karma.conf.js` via `process.env`.

Also, since I used async/await syntax which is recently included in ES2017, I made some changes to `.babelrc`, `gulpfile.babel.js`, `npm-shrinkwrap.json`, and `package.json`.

Is it ok to merge this PR?
If so, I will resolve conflicts with #1659.


asial-matagawa:
It seems that it does not work on CircleCI. I will fix this later.

asial-matagawa:
I thought `npm shrinkwrap` is the correct command to generate `npm-shrinkwrap.json` but actually it was incorrect.
I have generated `npm-shrinkwrap.json` with `npm shrinkwrap --dev` and it will work fine on CircleCI.

gulpfile.babel.js
```diff
@@ -135,6 +138,75 @@ gulp.task('core-test', ['prepare', 'core', 'core-dts-test'], (done) => {
...
+      listOfSpecFiles = glob.sync(path.join(__dirname, 'core/src/**/*.spec.js'));
+    }
+
+    // Separately launch Karma server for each spec file
+    try {
+      for (let i = 0 ; i < listOfSpecFiles.length ; i++) {
+        $.util.log($.util.colors.blue(path.relative(__dirname, listOfSpecFiles[i])));
+
+        // Pass parameters to Karma config file via process.env
+        process.env.SPEC_FILES_SPECIFIED = true;
```
argelius: Is there another way to pass these variables to Karma instead of changing `process.env`. I feel you should not change `process.env` inside a program since they are defined in the shell.

asial-matagawa: @argelius 
I found that we can use `global` instead of `process.env` and it also worked.
I have fixed `gulpfile.babel.js` and `karma.conf.js` to use `global`.

asial-matagawa:
@anatoo @argelius @frankdiox
I have additionally made the following changes:
- Use `global` instead of `process.env`.
- Add `--separately` option to `core-test` and remove `core-test-separately`.
- Rename the option `--target` to `--specs`.
- Remove unnecessary `global.SPEC_FILES_SPECIFIED`.

I think this PR could be useful when we fix the numerous problems which occur on browsers with WebReflection's/Polymer's CE0 polyfills.

Screenshot:


If there is no problem, could anyone merge this PR?

argelius:
Thank you @asial-matagawa ! I will merge this now.

PR #1669, opened 2016-11-01
fix(dialogs): event.cancel() reset.

frankdiox:
@argelius This fixes https://community.onsen.io/topic/983/prehide-event-on-ons-dialog-with-cancelable-attribute

Basically `this._running` isn't set back to `false` when `event.cancel()` is called.


argelius:
Thanks @frankdiox ! :rocket:

PR #1671, opened 2016-11-01
Pageloader refactor

frankdiox:
@argelius @anatoo This PR removes `rewritables.link` and uses pageLoader instead. I made a small refactor to pageLoader in order to make it work with AngularJS as well. Basically, the pages that were directly in the DOM from the beginning didn't have `unload` property. Now it doesn't matter because it's the pageLoader the one that knows how to unload pages.

@anatoo Is there any way to get `pageComponentRef` just from a DOM element? Looks like before there was `AppViewManager` to do this but it was removed after beta.16. I'm just temporarily saving `pageComponentRef` within the element in order to destroy it afterwards.

About `rewritables.ready`, looks like using  `setImmediate` or `setTimeout` instead works well but there is a problem with the e2e tests. I will look at it again but that can be done in a different PR.


anatoo:
@frankdiox Nice work!

> @anatoo Is there any way to get pageComponentRef just from a DOM element? Looks like before there was AppViewManager to do this but it was removed after beta.16. 

I investigated the way previously but probably there is nothing.

> I'm just temporarily saving pageComponentRef within the element in order to destroy it afterwards.

I think we should use `WeakSet` instead of saving directly within element.

bindings/angular2/src/directives/ons-navigator.ts
```diff
@@ -89,24 +89,32 @@ export class OnsNavigator implements OnDestroy {
...
+          {provide: OnsNavigator, useValue: this}
+        ], this._injector);
 
-      const factory = this._resolver.resolveComponentFactory(page);
-      const pageComponentRef = this._viewContainer.createComponent(factory, 0, injector);
-      const pageElement = pageComponentRef.location.nativeElement;
+        const factory = this._resolver.resolveComponentFactory(page);
+        const pageComponentRef = this._viewContainer.createComponent(factory, 0, injector);
+        const pageElement = pageComponentRef.location.nativeElement;
+        console.log(pageElement.data)
```
argelius: There is a console log here.

frankdiox: Nice catch :)

frankdiox:
@anatoo 
> I think we should use WeakSet instead of saving directly within element.

I guess you mean `WeakMap`. I actually thought about using that so I can use the pageElement as a key, but I didn't know if we can combine it with TypeScript and without polyfills. Do you think it's fine?

anatoo:
Oh, my bad. I mean `WeakMap` not `WeakSet`.

It should be fine to use WeakMap in TypeScript because there is `WeakMap` type declaration  in `bindings/angular2/typings/globals/es6-shim/index.d.ts`.

argelius:
@frankdiox I think it looks good!

PR #1675, opened 2016-11-04
fix(ons-switch): Fix `disbled` to `disabled`.

asial-matagawa:
@argelius @anatoo @frankdiox

I found a typo in `ons-switch.js`.
It might have been caused some issues until now.

If there is no problem, could anyone merge this PR?


argelius:
Thanks!

PR #1676, opened 2016-11-04
Fixed #1654.

### LINKED ISSUE #1654: ons-tab defaults to ion-cloud icon after pushPage()
**Environment**
Onsen 2.0.3
Angular 1.5.8
Safari, iOS

**Encountered problem**
When using an ons-tabbar and ons-tab elements in a template which opens after a navigator pushPage, all icons default to ion-cloud.

**How to reproduce**
Please check enclosed example
[Archief.zip](https://github.com/OnsenUI/OnsenUI/files/552971/Archief.zip)

I've added a pen: https://codepen.io/vjoomens/details/JRVbZR/
.

### LINKED ISSUE #1672: Receiving a lot of errors for `stopAll`
__Environment__
Onsen 2.0.3
iOS 10.x.x
WKWebView

__Encountered problem__
Using Track.js, I am seeing a lot of `undefined is not an object (evaluating 'n.stopAll')` apparently in iOS 10.x. The "undefined" in this case seems to be `this.animator`.

The application is running inside a WKWebView

Is this something known?

__How to reproduce__
I don't know, at the moment

anatoo:
@argelius I investigated #1654  and why `ons-icon` attribute change does not trigger `attributeChangedCallback()` in Safari. But I gave up and added dirty fix. Is there something that is considered to be the cause of that?

### DIFF
--- a/core/src/elements/ons-tab.js
+++ b/core/src/elements/ons-tab.js
@@ -272,7 +272,11 @@ export default class TabElement extends BaseElement {
     const badge = this.getAttribute('badge');
 
     if (typeof icon === 'string') {
-      getIconElement().setAttribute('icon', icon);
+      const iconElement = getIconElement();
+      const last = iconElement.getAttribute('icon');
+      iconElement.setAttribute('icon', icon);
+      // dirty fix for https://github.com/OnsenUI/OnsenUI/issues/1654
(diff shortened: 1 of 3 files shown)


argelius:
I think this fix is fine.

PR #1683, opened 2016-11-09
chore(angular2-binding): Updated angular.

anatoo:
@argelius Could you merge this changes if there is no problem?


argelius:
Thanks @anatoo!

PR #1685, opened 2016-11-09
Changed to include type definition file in Angular2 binding.

anatoo:
Summary:

 * Changed to reference "onsenui.d.ts" and removed unneeded "declare var ons: any;".
 * Changed to include "onsenui.d.ts" in angular2-binding distribution.

@argelius Could you merge this changes if there is no problem?


argelius:
Thanks @anatoo !

PR #1689, opened 2016-11-11
Update template and add more items into `Environment` section.

asial-matagawa:
@argelius @frankdiox @anatoo 
Since `ISSUE_TEMPLATE.md` is so outdated, I have updated it.
Also, I have made a major changes on `Environment` section so that users can easily and precisely describe their environment.

Since this PR is just a proposal, if anyone feels this new template too redundant I will discard this PR and update the template in the same way as before.
What do you think of this proposal?


argelius:
Thanks @asial-matagawa ! Are you willing to maintain this file so it doesn't get outdated?

asial-matagawa:
@argelius
Yes, I will update this as necessary.
Thank you for merging!

PR #1695, opened 2016-11-14
Added select CSS component

FREEZX:
I added a select component to replace the plain non-styled html component, achieved by masking an actual select component.
This is something i did quickly for my own needs and it needs review and guidance towards making it a part of OnsenUI.

### DIFF
--- a/css-components/components-src/stylus/components/index.styl
+++ b/css-components/components-src/stylus/components/index.styl
@@ -15,6 +15,7 @@
 @import './list'
 @import './search-input'
 @import './text-input'
+@import './select'
 @import './textarea'
 @import './dialog'
 @import './alert-dialog'
--- a/css-components/components-src/stylus/components/select.styl
+++ b/css-components/components-src/stylus/components/select.styl
@@ -1,0 +1,237 @@
+/**
+*
+* Copyright 2012 Adobe Systems Inc.;
+*
+* Licensed under the Apache License, Version 2.0 (the "License");
+* you may not use this file except in compliance with the License.
+* You may obtain a copy of the License at
+*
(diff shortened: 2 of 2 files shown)


argelius:
@FREEZX Thank you for the PR! It's really appreciated. A select component is something that we would like to add.

I think there are some issues with this implementation. I will try to list them here.

* Firstly, I think it is broken because it will not display the currently selected item.
```html
<div class="selected-option">
  Selected <!-- It always displays select, we need to use the native select element to display the correct value. -->
</div>
```
* The structure is too complicated, this requires the developer to create a lot of elements. I think it can be simplified so you don't need to use so many DIVs. It should be possible to just style the `<select>` element itself.
* There is no need to commit the generated files. Please only commit the `.styl` files since that makes it easier to read the diff.

I think a select component should look almost exactly like a normal text input in Material Design but with a caret icon. There is an example here: https://material.google.com/components/menus.html#menus-behavior

FREEZX:
@argelius Thanks for the feedback, i'll try to fix the issues.

FREEZX:
There we go, seems like just applying the input style did the trick

frankdiox:
@FREEZX Hey! Sorry for the late reply. I think this is now the correct approach, thanks. Right now we don't have a way to document CSS classes so I'd like to make a component that uses this internally with autostyling feature. We will address this when we finish some other stuff we are doing :)

frankdiox:
@FREEZX Hey! We made an `ons-select` component with your Stylus code. I've merged it into another branch and will be released soon. Therefore, I'm closing this now. Thanks!

FREEZX:
Awesome! Thanks!

PR #1701, opened 2016-11-16
Input style

frankdiox:
@argelius `transparent` modifier and a simple fix for `float` attribute style.


argelius:
Thanks @frankdiox!

PR #1710, opened 2016-11-22
Fix angular2 navigator init

argelius:
@anatoo @asial-matagawa 

I think we need to force change detection when loading a page for the navigator. Otherwise, the view won't be updated. I had an issue where the first page was not displayed correctly and this changed fixed it. 

Maybe it's because the first page is pushed outside the zone. :question: 

EDIT: Actually, with this fix I am still having an issue with change detection not happening inside asynchronous callbacks. Zone.js is supposed to take care of that but maybe it doesn't work in dynamically created components?


argelius:
I will close this since it seems like it is breaking Zone.js

PR #1723, opened 2016-11-30
feat(react): add react bindings

argelius:
@frankdiox 

This adds React bindings to the main repo. I will add a note to the original repo when this has been merged.

Please don't merge this yet. I will also need to remove the submodule from the website repo and use this instead.


argelius:
Actually, I guess this can be merged regardless. It just needs to be merged and a new version tagged before https://github.com/OnsenUI/onsen.io/pull/303 is merged.

frankdiox:
@argelius Just one thing. What about the redux navigator PR in the other repo?

argelius:
@frankdiox I forgot about that. I will do that as a separate task.

I need to document it and I would also like to update the Redux weather app to use it.

I will add this task to the spreadsheet.

PR #1724, opened 2016-12-01
fix(angular1):  is now the original Custom Event.

frankdiox:


argelius:
I think this is a good fix

PR #1741, opened 2016-12-08
React new navigator

argelius:
@frankdiox 

This implements a stateless navigator. I have provided an example using Redux. As we discussed before, our library might be able to implement a higher order component to make this easier to use with Redux.


frankdiox:
@argelius Thank you! The issues have been fixed :)

Issue #1711, opened 2016-11-24
`class` attribute not working for `<ons-button>` in Angular 2


argelius:
[](Github issues should be used only for bugs and feature requests. For questions please visit https://stackoverflow.com/tags/onsen-ui or https://community.onsen.io. For bugs please use the following template.)

__Environment__ 

```
[Core]
    Onsen UI 2.0.4

[Framework]
    Angular 2.1.2

[Framework binding]
    angular2-onsenui 1.0.0-rc.1

[Additional libraries]

[Platform]
    Desktop - Linux

[Browser]
    Desktop - Chrome 54.0.2840.71
```

__Encountered problem__

When using the `class` attribute the `button` class is removed from the element.

__How to reproduce__

Create an `<ons-button>` with a class:

```
<ons-button class="my-class">My button</ons-button>
```

anatoo:
This is not bug and correct behavior. Because Angular 2 set `my-class` value to `class` attribute **after** creating `<ons-button>` element and add "button" to `classList` property.

We must change spec of "class" attribute behavior to fix this problem.

PR #1753, opened 2016-12-15
Changed class attribute behavior.

anatoo:
@frankdiox I changed class attribute to contain default class token always on some elements. 

This fix https://github.com/OnsenUI/OnsenUI/issues/1711. Could you merge this changes if there is no problems?


frankdiox:
@anatoo Thanks for the fix and prettifying the if-else :)

PR #1759, opened 2016-12-22
Lazy repeat dynamic heights

frankdiox:
@anatoo @asial-matagawa Here's an improvement for lazy-repeat.

Previously I made a very different version where new items were added synchronously for scroll up and asynchronously for scroll down (due to Angular 2 bindings). That one was working and did not require anything like `calculateItemHeight`, but the code is quite complex since it needs to handle race conditions and many other stuff.

While doing that I came up with a different approach that requires `calculateItemHeight` only as an approximation in order to know how many extra items it should render. However, after the rendering is done it takes the actual height of every element and uses it to keep the proper scroll and list height. This is not using absolute position anymore, everything is a block and it's pushed down by `ons-list`'s paddingTop. This way users can do things like "opening" a visible item (growing its height) without any problem. The current limitation is that we cannot modify the height of the non rendered items at the top of the list without having a small jump in the scroll, but I don't think this is very important.

I just want your opinion about this approach and maybe a quick review if you guys have time. We don't need to merge this right now.


asial-matagawa:
@frankdiox 
Sorry, I have not gone deep into the implementation of lazy repeat, so I cannot tell you anything.

@anatoo 
Could you review this commits?

anatoo:
@frankdiox I read all your changes and tried `lazy-repeat` examples on several browsers. It looks good approach to me. Great work! 

Just one thing,  I thought the scrolling performance is a bit poor when I tried on mobile Safari. So I profiled on chrome devtools and it seems that some forced layouts cause this problem. 
image

Could you try to improve scroll performance on `lazy-repeat`?

frankdiox:
@anatoo Thanks! Yeah sure, I haven't slowly tested performance or memory leaks yet, I'll do that before merging :)

frankdiox:
@anatoo I think the reflows are now gone. At least in Chrome browser I cannot see any related problem anymore. Only with Angular1 bindings I can see some reflows due to Angular itself invalidating the layout. In any case, this also happened in the old implementation of lazy repeat so it shouldn't be a problem.

Could you please check it on mobile Safari once again? Thanks!

anatoo:
@frankdiox I checked on mobile Safari and it seems that the scroll performance is improved. Good!
But the `lazy-repeat` behavior is still unstable on Safari and mobile Safari. I recorded that on webm: [lazy-repeat.webm.zip](https://github.com/OnsenUI/OnsenUI/files/700947/lazy-repeat.webm.zip).
Could you try to fix this unstable behavior?

frankdiox:
@anatoo Oh I see, thanks! I guess I should wait until I'm in the office to test this in Safari...

anatoo:
@frankdiox OK.

frankdiox:
@anatoo I've tested this in MacOS Safari and iOS Safari and can't break it anymore. And no reflows, I think it should be good to go :slightly_smiling_face:

anatoo:
@frankdiox Great work! I will merge this changes.

PR #1765, opened 2017-01-05
Fix 1726.

anatoo:
@asial-matagawa @frankdiox  To fix #7126, I removed `onsPage.name` and adopted `navigator._pageMap` WeakMap property instead. Could you merge this changes if there is no problem?


frankdiox:
@anatoo Thanks! :+1:

PR #1779, opened 2017-01-14
chore(react-binding): Replace string refs with callbacks.

matoilic:
String refs are considered legacy and are likely to be removed in one of the future releases of React (see https://facebook.github.io/react/docs/refs-and-the-dom.html#the-ref-callback-attribute). This pull request replaces all string refs with callbacks.


frankdiox:
@matoilic Thanks a lot! I will test this on Monday and merge it!

PR #1784, opened 2017-01-18
Vue2 test

frankdiox:
@asial-matagawa This continues Vue 2 implementation. I think you can check commits one by one, they are well separated.

Some comments:
 * [feat(vue): VuePageLoader inherits props.](https://github.com/OnsenUI/OnsenUI/commit/35ab81af414e3dac02129fee0858c44b6ff27b45): This is just a test for now. The props are passed but primitive value props are not reactive. Somehow I think we need to pass the props down, perhaps we can find a better way. I think we should support the "core API" like we do right now (with `page` attribute, etc) but also implement some Vue specific API where we can directly change the child pages.
 * [feat(vue): Added generic components.](https://github.com/OnsenUI/OnsenUI/commit/e380b5e3da88454806a2532e6b8dc9a1f298d22c): I've made a generic way to create components instead of having many files. Do you think this is fine?

I'll keep working on this branch.


bindings/vue/src/internal/optionsObjectHelper.js
```diff
@@ -58,7 +59,7 @@ const createComputedPropertiesFor = (targetClass) => {
       // register a computed property
       // which relay set/get operations to its corresponding property of DOM element
       computed[propertyName] = {
-        get() { return this.$el[propertyName]; },
+        get() { return this.$el[propertyName].__vue__ || this.$el[propertyName]; },
```
asial-matagawa: `this.$el[propertyName].__vue__` throws an exception if `this.$el[propertyName]` is null. This should be fixed

frankdiox: That was fixed in my local branch, I forgot to push.

Issue #1801, opened 2017-02-03
ons-splitter in ons-navigator bug on Safari


puku0x:
__Environment__ 

```
[Core]
    Onsen UI 2.1.0

[Framework]
    Angular 1.5.9

[Framework binding]
    angular-onsenui 2.1.0

[Additional libraries]
    None

[Platform]
    Mobile - iOS 10.1.1
    Desktop - macOS Sierra

[Browser]
    Mobile - iOS UIWebView
    Mobile - iOS Safari
    Desktop - Safari 10.0 (12602.1.50.0.10)
```

__Encountered problem__

`<ons-splitter-content>` and `<ons-splitter-side>` are broken when they are in `<ons-navigator>` .

__How to reproduce__

Please refer to this.
http://codepen.io/puku0x/pen/JEvpQP

asial-matagawa:
@puku0x 
Thank you for reporting.

Actually I think this is not a bug since `ons-template` cannot be nested, that is, `ons-template` elements cannot include child `ons-template` elements:

>However, having multiple pages declared in the same file is sometimes more convenient.
ons-template component is the solution for this.
>You can create as many pages as you need **in your index.html (there and only there)** by using this component.

https://onsen.io/v2/docs/guide/js/#templates

Once you move inner `ons-template` elements to outside of the root `ons-template`, it will work fine.

But we should warn a user when `ons-template` is located in `ons-template`, so I will add `console.warn` to `ons-template` in the core.
Thank you!

PR #1802, opened 2017-02-03
Show warning when ons-template is not located just under document.body.

asial-matagawa:
@anatoo @frankdiox 

This PR enables `ons-template` to detect nested `ons-template` elements.
Nested `ons-template` elements cause unexpected behaviors like #1801, so I think we should show warning when they are used.

This feature works like:

Is there any problem with this?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 CHANGELOG
 ====
 
+v2.1.1
+----
+ * ons-template: Show warning when ons-template is nested.
+
 v2.1.0
 ----
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ CHANGELOG
 
 v2.1.1
 ----
- * ons-template: Show warning when ons-template is nested.
+ * ons-template: Show warning when ons-template is not located just under document.body.
 
 v2.1.0
 ----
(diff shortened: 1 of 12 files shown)


frankdiox:
@asial-matagawa Perhaps we should check that the parent is `document.body`? This way we show the warning if the template is not located in `index.html` under `<body>` tag directly. I think that's the desired behavior.

asial-matagawa:
@frankdiox 
>Perhaps we should check that the parent is document.body?

That is better and I have fixed my implementation according to your suggestion.
The new implementation shows the parent's outerHTML.

Thank you!

asial-matagawa:
@frankdiox 
Maybe completely fixed now.

`this.parentNode` can be `null` even after `connectedCallback` is called due to some CE polyfill-related bug, so I have added `if (this.parentNode)`.

frankdiox:
@asial-matagawa Looks like it still fails in Safari. Can you have a look at it?

asial-matagawa:
@frankdiox 
I think this PR is now safe for merging.
Could you merge this?

frankdiox:
@asial-matagawa Thanks! In any case, as I mentioned the other day, perhaps we should look into native `template` and perhaps deprecate `ons-template`. Looks like @anatoo [also thinks alike](https://twitter.com/anatoo/status/831099657048645632) :)

PR #1804, opened 2017-02-06
Angular ons-select

misterjunio:
@frankdiox @asial-matagawa `ons-select` wrappers for Angular1 and Angular2 plus examples.


frankdiox:
@misterjunio Thanks a lot! Few things:
  - Do you think we need `var` attribute in `ons-select` for AngularJS? We have it for `ons-switch` but not for `ons-input`. Basically, `var` avoids the need of accessing the DOM element (what should not be necessary in AngularJS). If you think `ng-model` and `ng-repeat` is enough then it's fine.
  - Could you add docs in `directives/ons-select.js`? All the other components documents `on`, `off`, `once`, etc.
  - Could you modify the core `ons-select` and add support for an existing `select` element inside `ons-select`? This would allow users to use `ng-options` if they really want it, among other things. This can be done in a different PR if you prefer, I don't mind.

misterjunio:
@frankdiox On it

misterjunio:
@frankdiox Regarding your points:

- I was checking and I don't think we need `var` for select. As you said `ng-model` already allows `ng-change` listeners and with `ng-options` users can populate the field so I don't predict they will need more DOM manipulation that justifies creating a special view for this element.

- I added the docs. Guess there are no other methods than those 3.

- Not sure what you mean. Do you want to allow them to create their own `select` (for instance with ng-options) inside `ons-select`? And in that case the purpose of `ons-select` would be basically just styling, is that it?

frankdiox:
@misterjunio Yes, I think either of these two templates should be valid:

```
<ons-select>
  <option></option>
  <option></option>
</ons-select>
```

```
<ons-select>
  <select>
    <option></option>
    <option></option>
  </select>
</ons-select>
```

So instead of creating the `select` element always, just check first if it exists and use it instead.

misterjunio:
@frankdiox Alright, I pushed those changes in the branch `fix-ons-select` 👍

frankdiox:
@misterjunio Great! I think you need to push the docs and changelog, though :sweat_smile:

misterjunio:
@frankdiox Of course 😦

frankdiox:
@misterjunio :+1: :+1: :hotsprings:

PR #1805, opened 2017-02-06
chore(*): Automatically build CSS components and live reload.

frankdiox:
@asial-matagawa @anatoo With the new changes in `css-components`, `gulp build` deletes the generated CSS components since it performs `gulp clean` as well. I added a task to build the CSS components directly with `gulp build` in the parent directory. Also, it now watches the generated `build/css/onsen-css-components.css` and reloads the browser. This is useful if we serve both Onsen UI and css-components.

I guess there is no problem with this changes but I just want to inform you guys about this "feature".


anatoo:
@frankdiox Thanks. Nice work!

PR #1818, opened 2017-02-16
style(ons-select): Fixed material variables

misterjunio:
@anatoo I defined some variables for material select-input but in fact I'm not an expert so I have no idea if they're ok, can you check?


anatoo:
@misterjunio It's OK. I will port these stylus variables to less components. Thanks!

misterjunio:
@anatoo Thank you 👍

PR #1822, opened 2017-02-16
fix(RouterNavigator): Fixes undefined variable error in RouterNavigator

caiobalthazar:
Attempts to fix:
```
[...]node_modules/react-onsenui/dist/react-onsenui.js:3849
Navigator.propTypes = {
^

ReferenceError: Navigator is not defined
```


frankdiox:
@caiobalthazar Oops, thanks!

Issue #1827, opened 2017-02-17
Input type number not working in AngularJS


nickpdiana:
__Environment__ 

```
[Core]
    Onsen UI 2.1.0

[Framework]
    Angular 1.5.9

[Framework binding]
    angular-onsenui 2.1.0

[Additional libraries]
    jQuery 3.1.1

[Platform]
    Desktop - macOS Sierra

[Browser]
    Desktop - Chrome 55.0.2883.75
```

__Encountered problem__
ons-input elements of type number do not retain that type

__How to reproduce__

I've attached a screenshot showing the issue. In the console, we see that ct.moneyInBank is initialized as a number, but once I edit the value in the ons-input box, it is converted to a string.

Here is the relevant code:
`<ons-input placeholder="Money in the Bank" type="number" float ng-model="ct.moneyInBank" ng-change="ct.drawLine()"></p>`

Here is a screenshot:
<img width="911" alt="screen shot 2017-02-17 at 6 01 26 pm" src="https://cloud.githubusercontent.com/assets/2049108/23086592/dfdb9950-f53b-11e6-8cdb-ac6bf54f308f.png">

misterjunio:
@nickpdiana We treat inputs different than radio buttons, checkboxes, selects and ranges as text inputs. Is there a specific reason why you need the type of the variable to be `number`?

nickpdiana:
Just for the same reasons why [vanilla input tags](http://w3c.github.io/html-reference/input.number.html) have a type number:

1) I don't have to cast the variable to an integer anytime I use it.
2) Mobile users are given a number pad for input instead of a keyboard.

My current solution is to just use a vanilla input tag styled with onsen classes, but that's not ideal.

misterjunio:
I think you may be a bit confused about what the input type number means because even a vanilla input will give you a string variable. From the link you posted:

> type = "number"
Specifies that its input element is a precise control for setting the **element’s value to a string representing a number**.

You can also try to log the `typeof` the element's value and it should give you `string`.
About the second point I'm not sure I am able to reproduce because both with and without `ons-` I get the keyboard but on the numbers and symbols pad which is the expected behavior.

nickpdiana:
Because of the data-binding in angular, the value for input[number] elements must be of type number. Here's a [jsbin](https://jsbin.com/mucewutuvu/edit?html,js,console,output).

[...]

PR #1828, opened 2017-02-21
Angular1 input fix

misterjunio:
@frankdiox @asial-matagawa Small fix for issue #1827, I'll merge it to master if you have no problem.

### DIFF
--- a/bindings/angular1/directives/input.js
+++ b/bindings/angular1/directives/input.js
@@ -69,11 +69,7 @@
           const set = $parse(attrs.ngModel).assign;
 
           if (el._isTextInput) {
-            if (el.getAttribute('type') == 'number') {
-              set(scope, Number(el.value));
-            } else {
-              set(scope, el.value);
-            }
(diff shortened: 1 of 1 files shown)


frankdiox:
@misterjunio Thanks! Just minor things:
1. Please use the strict equality operator (`===`) instead of `==` just to be consistent with the styles.
2. `type="NUMBER"` will not meet the equality because of the letter case but it will still be considered as a number input. You can either use `toLowerCase` method or even better, just check the `type` property instead of the attribute if it's available. If the attribute `type` is omitted, `getAttribute` returns `null` but the property `type` falls back to its default value `text`.

```
<input type="NUMBER">

>> $('input').type
>> "number"
>> $('input').getAttribute('type')
>> "NUMBER"
```

misterjunio:
@frankdiox Fancy 😄  should be good to merge now 👍

frankdiox:
@misterjunio Thanks!

PR #1832, opened 2017-02-22
Babelrc

frankdiox:
@asial-matagawa Can you have a look at this? We should be a bit more consistent with the presets across the bindings. This adds `latest` and `stage-3` as presets.


asial-matagawa:
@frankdiox 
I think it is fine!

BTW, the reason why you used `stage-3` is because we can notice when `transform-object-rest-spread` is moved to `stage-4`, right?

frankdiox:
@asial-matagawa [stage-3](https://babeljs.io/docs/plugins/preset-stage-3/) should be safe to use and it includes `transform-object-rest-spread`, yes.

asial-matagawa:
@frankdiox @anatoo
>By the way, Angular2 bindings only use es2015 but I didn't want to modify that. You guys can update it if you want.

Sure, thank you!

PR #1834, opened 2017-02-24
Dropped less files and rebuild css-components with CSSNext.

anatoo:
@misterjunio @frankdiox @asial-matagawa I reimplemented css-components with CSSNext. I will merge this changes if there is no problems.


frankdiox:
@anatoo Awesome, thanks!

PR #1835, opened 2017-02-24
fix(ons-popover): Device back button behavior.

frankdiox:
@anatoo I think we can fix the deviceBackButton behavior in this way. Now the tree only contains the nodes that can be used for the comparison, not always the parents. It's the first time I modify this algorithm so please have a quick look at the changes.

The tree with navigator + popover would be: `[ons-navigator, div.popover]`. Before it was `[ons-navigator, ons-popover]`. And if the popover was located inside a span, for example, it was `[ons-navigator, span]`.

Do you think this is fine or it could break something?


anatoo:
@frankdiox Thanks! It should be fine.

PR #1840, opened 2017-02-28
Refactored progress-circular CSS components.

anatoo:
Simplified CSS progress-circular components. I will merge this change to master branch.


asial-matagawa:
I have reopened #1860 to deal with the problem you reported.

s-nel:
Can we revert `cx`, `cy`, and `r` changes to fix issues in Firefox and iOS 8?

PR #1850, opened 2017-03-06
feat ons-navigator: add get listener on device back button dispatcher

francesco-strazzullo:
I Added a getter for the device back button handler. It could be useful for reset the listener to the original one. I'm using it in a React Application with this use case: In a Page I want a different back management, but after changing the page I want to reset the default behavior.

### DIFF
--- a/core/src/ons/device-back-button-dispatcher.js
+++ b/core/src/ons/device-back-button-dispatcher.js
@@ -178,6 +178,10 @@ class DeviceBackButtonDispatcher {
         this._callback = callback;
       },
 
+      getListener: function() {
+        return this._callback;
+      },
+
       enable: function() {
         HandlerRepository.set(element, this);
       },
--- a/core/src/ons/ons.js
+++ b/core/src/ons/ons.js
@@ -110,6 +110,16 @@ ons.ready = callback => {
 };
 
 /**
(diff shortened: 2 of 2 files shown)


frandiox:
@francesco-strazzullo Hi, thanks for the PR! This is just the "master" handler, which is always `navigator.app.exitApp()`, so you could set it back by just writing that line. I think what you really want to set is a handler for some specific component like Navigator, Dialog, Page, etc. Am I wrong?

For example, [this](https://onsen.io/v2/docs/js/ons-page.html#properties) is the core `ons-page` which have a `onDeviceBackButton` property. More info about components handlers [here](https://onsen.io/v2/docs/guide/js/#device-back-button).

**Looks like there is no device back button handler prop in React Components** so I think that's the place where we should fix this. We should have a prop `onDeviceBackButton` or something similar in the React components.
Could you try to modify that instead ([`bindings/react/src/components`](https://github.com/OnsenUI/OnsenUI/tree/master/bindings/react/src/components))? Otherwise, if you can access the DOM element (with refs) you can set the property there as a quick dirty fix.

francesco-strazzullo:
@frandiox Ok, it seems cleaner in that way. I will try that.

PR #1851, opened 2017-03-06
feat: react binding page - device back button

francesco-strazzullo:
I Added an onDeviceBackButton to wrap the core page 'onDeviceBackButton'. It's related to the #1850

### DIFF
--- a/bindings/react/src/components/Page.jsx
+++ b/bindings/react/src/components/Page.jsx
@@ -40,7 +40,6 @@ class Page extends BasicComponent {
   }
 
   componentWillReceiveProps(nextProps) {
-    super.componentWillReceiveProps(nextProps);
     var node = ReactDOM.findDOMNode(this);
     node.onDeviceBackButton = nextProps.onDeviceBackButton;
   }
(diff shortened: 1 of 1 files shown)


frandiox:
@francesco-strazzullo Thanks! A couple of things: I think we don't need 72652b5 anymore, right? Also, according to React docs:

>React doesn't call componentWillReceiveProps with initial props during mounting. It only calls this method if some of component's props may update.

Therefore, I guess it doesn't take the initial value. Am I wrong? We are setting a similar prop in [`componentDidMount` for PullHook](https://github.com/OnsenUI/OnsenUI/blob/master/bindings/react/src/components/PullHook.jsx#L41). I guess that means it is "init only" value, but that's probably fine. Can you have a look at this?

Thanks a lot!

francesco-strazzullo:
@frandiox Yeah you can close the other PR. I think that it's not useful anymore. I think that it's ok to set this value in ```componentDidMount``. I will change it right away.

francesco-strazzullo:
@frandiox it's done!

frandiox:
@francesco-strazzullo Thanks! I'll do this for the other components as well before I forget :)

Issue #1758, opened 2016-12-21
React: Dialogs are created outside of the root context


corneyl:
Because [OnsenIO/react-onsenui](/OnsenUI/react-onsenui) is closing down, here a repost of the issue.

__Environment__
+-- onsenui@2.0.4
+-- react@15.4.0
`-- react-onsenui@1.0.3

It would be nice in my opinion to create all `Dialog`s (and `AlertDialog`s etc.) within the `Navigator` or root context.

The problem now is that when one uses a root tag which provides a certain context in React, the `Dialog`s do not have this context. Examples are the context provided by `MuiThemeProvider` ([link](http://www.material-ui.com/#/get-started/usage)) or `react-intl`'s `IntlProvider` ([link](https://github.com/yahoo/react-intl)) from which the theme and i18n functionality cannot be used in `Dialog`s etc. because they lack the proper context.

It would be nice if there is a solution for this problem, but maybe it's not possible to fix this easily.

anatoo:
@corneyl Thanks for reporting. Okay, we will try to fix this problem.

frandiox:
Fixed in #1852

PR #1852, opened 2017-03-06
fix(react): Fix broken context in dialogs (#1758)

matoilic:
This PR fixes the issue mentioned in #1758, where elements within dialogs have a different context than the React application itself. 

`renderSubtreeIntoContainer` is currently marked as unstable because the API might change in the future but it will not be removed any time soon. This is according to one of the core team members  https://twitter.com/soprano/status/737316379712331776. This is currently the only way I know of how to render elements outside the application root and to preserve the context at the same time.


frandiox:
@matoilic Thanks a lot! I'm OK with this, I'd just like to ping @anatoo first.

anatoo:
@matoilic Thanks! I will merge this.

PR #1853, opened 2017-03-07
bug

net900621:
Vue.util has been already canceled，so we need to add the 'hyphenate' function.


asial-matagawa:
@net900621 
Thank you for PR, but it has been already fixed in another branch (`vue-api`):
https://github.com/OnsenUI/OnsenUI/pull/1837/files#diff-cc347eda69595b59d60fa72f905de0b8

We will soon release `vue-onsenui 2.0.0-alpha.1` or `vue-onsenui 2.0.0-beta.0` with the fix. Thank you!

net900621:
thx

PR #1861, opened 2017-03-09
Fix #1768.

### LINKED ISSUE #1768: React : Some callbacks (onChange and OnClose) are only treated once
__Environment__ 

```
[Core]
    Onsen UI 2.0.4

[Framework]
    React 15.4.1

[Framework binding]
    react-onsenui 1.0.4

[Additional libraries]
    None

[Platform]
    Desktop - Windows 10

[Browser]
    Desktop - Chrome 55.0.2883.75
```

__Encountered problem__

Some callbacks like `Switch.onChange` or `SplitterSide.onClose` don't use updated prop value. If a new function is passed to the component in a render cycle, the original function still is used.

This is a big issue with _functional stateless component_ because they rely on closures, not on `this.state` or `this.props`.

__How to reproduce__

Here is a codepen that demonstrate the issue : https://codepen.io/anon/pen/NdPzJa?editors=1011

- The outer container (`MyPage`) uses `setInterval` to increment  `i`
- `MyPage` contains a `MyComponent` which is a _functional stateless component_
- `MyComponent` uses `i` in two different callbacks : a callback to `Switch.onChange` and a callback to `Button.onClick`
- When you click the button, the correct value is displayed in the console.
- Bug : When you click the Switch, the first value (`1`) is displayed in the console.

anatoo:
Fix #1768.

@asial-matagawa @frandiox Can I merge this changes to master branch now?

### DIFF
--- a/bindings/react/CHANGELOG.md
+++ b/bindings/react/CHANGELOG.md
@@ -4,6 +4,7 @@ CHANGELOG
 
 dev
 ----
+* Fixed [#1768](https://github.com/OnsenUI/OnsenUI/pull/1768).
 * router-naigator: Fixed [#1822](https://github.com/OnsenUI/OnsenUI/pull/1822).
 * lazy-list: Support for dynamic heights.
 * select: Add Select component.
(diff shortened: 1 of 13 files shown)


asial-matagawa:
@anatoo
`bindings/react` directory does never affect `css-components`, `core` and `angular-onsenui`, so I think this PR can be merged.
Maybe we should think about production branches for each binding...

@frandiox 
What do you think?

frandiox:
@anatoo Thanks for the fix! :+1:

PR #1865, opened 2017-03-09
fix(examples): Small fix for two examples.

misterjunio:
Small improvement of two core examples. I will merge it if there are no problems.

### DIFF
--- a/examples/gesture_detector/index.html
+++ b/examples/gesture_detector/index.html
@@ -1,5 +1,6 @@
 <!doctype html>
 <html lang="en">
+
 <head>
   <meta charset="utf-8">
   <title>Gesture Detector | Onsen UI</title>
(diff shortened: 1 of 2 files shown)


frandiox:
@misterjunio That Lazy Repeat example looks exactly like the old one before I modified it :sweat_smile: 
The change I made is for checking that Lazy Repeat works with dynamic heights. Do you think we don't need that?

misterjunio:
@frandiox Whaaat I didn't notice you had changed it, I thought I was only removing the `setup` thing, I'm sorry. I just pushed a version with just the gesture detection change. If it's ok please merge it

frandiox:
@misterjunio Don't worry! I know it's ugly code but it's useful for testing :sweat_smile:

PR #1870, opened 2017-03-13
Fix broken slide animation in iOS 9 and iOS 10.

asial-matagawa:
@frandiox
In iOS 9 and iOS 10, when translate3d animation starts from outside of viewport, the element immediately warps to the destination (maybe a bug of iOS Safari).
To prevent that, we have to put a vertically long element into ons-dialog element.

This PR fixes it by putting a vertically long element into .dialog element.
This solution worked in both of iOS 9 and iOS 10.

Could you review this?


asial-matagawa:
Additional tasks have been done. I will merge this.

PR #1885, opened 2017-03-23
fix(css-components): Fixed list-header

puku0x:
Thank you Onsen UI team !

I found some bug in `css-components/src/components/list.css`.
- `--list-header-background-color` was not defined
- Text color has no opacity
- There is a magic value of `#6c6c6c`

I made a PR, please check it.

image
Left: iOS, Right: Onsen UI v2.2.0 (the current version is not alike iOS)


asial-matagawa:
Thank you puku0x!

@anatoo
Could you take a look at this?

anatoo:
@puku0x Thanks! I will review this PR.

asial-matagawa:
@anatoo 
FYI: As we decided in the meeting 2 weeks ago, we should not immediately merge patches into `master` branch. We should release patches (and new `CHANGELOG` and `package.json`) by using `production` branch and after that we should merge the patches into `master`.

asial-matagawa:
Also, we should fork the current `production` branch (https://github.com/OnsenUI/OnsenUI/tree/production).
The current `master` branch contains dirty commits which should not be released now.

puku0x:
I found my PR is not enough to solve this problem, therefore I cancel it.
I will make another issue for `ons-list-header`.

PR #1886, opened 2017-03-23
Deep toolbar animation

frandiox:
@asial-matagawa @misterjunio Please merge if there's no problem :)

### DIFF
--- a/core/src/ons/util.js
+++ b/core/src/ons/util.js
@@ -228,6 +228,35 @@ util.findFromPath = (path) => {
 };
 
 /**
+ * @param {HTMLElement} container - Page or page-container that implements 'topPage'
+ * @return {HTMLElement|null} - Visible page element or null if not found.
+ */
+util.getTopPage = container => container && (container.tagName.toLowerCase() === 'ons-page' ? container : container.topPage) || null;
+
+/**
+ * @param {HTMLElement} container - Element where the search begins
+ * @return {HTMLElement|null} - Page element that contains the visible toolbar or null.
+ */
+util.findToolbarPage = container => {
+  const page = util.getTopPage(container);
+
+  if (page) {
+    if (util.findChild(page, 'ons-toolbar')) {
+      return page;
+    }
+
+    for (let i = 0; i < page._contentElement.children.length; i++) {
+      const nextPage = util.getTopPage(page._contentElement.children[i]);
+      if (nextPage) {
+        return util.findToolbarPage(nextPage);
+      }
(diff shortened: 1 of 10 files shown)


core/src/ons/util.js
```diff
@@ -228,6 +228,35 @@ util.findFromPath = (path) => {
...
+util.findToolbarPage = container => {
+  const page = util.getTopPage(container);
+
+  if (page) {
+    if (util.findChild(page, 'ons-toolbar')) {
+      return page;
+    }
+
+    for (let i = 0; i < page._contentElement.children.length; i++) {
+      const nextPage = util.getTopPage(page._contentElement.children[i]);
```
asial-matagawa: What about `innerPage` instead?

frandiox: If it's already a page it returns itself, so I thought "innerPage" could be misleading.

core/src/ons/util.js
```diff
@@ -228,6 +228,35 @@ util.findFromPath = (path) => {
 };
 
 /**
+ * @param {HTMLElement} container - Page or page-container that implements 'topPage'
+ * @return {HTMLElement|null} - Visible page element or null if not found.
+ */
+util.getTopPage = container => container && (container.tagName.toLowerCase() === 'ons-page' ? container : container.topPage) || null;
```
asial-matagawa: I think we can avoid creating public methods `ons-splitter#topPage` and `ons-toolbar#topPage` by replace `'ons-page' ? container : container.topPage` with `'ons-page' ? container : (container.topPage || container._topPage)`.
What do you think?

frandiox: We are already using `topPage` for the Navigator so I think using the same API makes sense. It is not documented for Splitter or Tabbar so we can change it at any time without problems. We could even document it since it could be useful sometimes...

asial-matagawa: (This discussion continues to our Slack channel and this problem was solved)

core/src/ons/util.js
asial-matagawa: I see, if so `page` is easy to read for me.
`nextPage` sounds related to navigators in this case.

asial-matagawa: Ah, `page` is already used...

frandiox: Well, it is basically the "next page to check".

asial-matagawa: Let's use `nextPage`. Sorry for interrupting.

PR #1887, opened 2017-03-23
fix device back button handler

airamrguez:
onDeviceBackButton is lost on Navigator component re-renders


frandiox:
@airamrguez Thanks for the PR! Is this also an issue in other components that support device back button (dialogs, splitter...)?

airamrguez:
I haven't tested them but I suppose so looking at the code. Back button handler is only attached within `componentDidMount`. Subsequent re-renders of the VDOM that attaches new handlers will also be ignored. For that reason it has to be reattached within `componentWillMount`.

PR #1888, opened 2017-03-27
fix(ons-fab): Page transition animation with fabs.

frandiox:
@asial-matagawa This was fixed some time ago but somehow it was changed again. This fixes again the transitions with FAB/SpeedDial components.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,8 +12,9 @@ dev
  * ons-navigator: iOS Slide animation can now find toolbars that are not immediate children of the pages.
  * ons-pull-hook: Its content is not visible anymore during toolbar transitions.
  * ons-fab: It is now initally hidden and shown only when its page container is pushed. This fixes page transitions.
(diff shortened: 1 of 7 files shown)


frandiox:
@asial-matagawa Some more fixes. Looks like the toolbar's zIndex was updated at some point but the fab's wasn't. Also, when popping pages in the navigator, the FAB/SpeedDial disappears suddenly without animation. It should be done asynchronously.

I removed 2 broken tests because I think they don't make sense, the show/hide should be tested in another way.

asial-matagawa:
@frandiox 
Thank you for the changes.
It could be said that `ons-fab` potentially depends on `ons-navigator` and other components...

frandiox:
@asial-matagawa It only depended on `ons-page` and its lifecycle. Anyway, I added an extra `show` that also fixes the animation so it doesn't directly depend on the page anymore for this. Should be fine now.

asial-matagawa:
I see, nice work!

PR #1889, opened 2017-03-27
Lock loading order of the polyfills and refactor the directory structure.

asial-matagawa:
@frandiox 
In the current implementation of Onsen UI, the loading order of the polyfills can vary depending on file paths of them.
So I have locked the loading order without any breaking changes.
(It generates the same `onsenui.js` as before)

Also, I have refactored the directory structure and added strict version information of the polyfills for better management (especially for managing Custom Elements polyfills).

If there is no problem, I would like to merge this into `master`.


frandiox:
@asial-matagawa Thanks! Is there any reason why we should have both `polyfill` and `polyfill/vendor`? Isn't it enough with `polyfills/` for example?

asial-matagawa:
@frandiox 
Sorry, I missed the notification...

I found that some files are not 3rd-party ones but our original files, so I placed `CustomEvent.js`, `childNodeRemove.js` and `viewport.js` outside the `vendor` directory. (FYI, `viewport.js` was not even a polyfill...)

frandiox:
I don't think it makes any difference, we don't really need to separate them. But anyway, I guess it's not that important. Thanks!

asial-matagawa:
I see, thank you!
I will remove `vendor` directory and use `polyfills` instead of `polyfill`.

asial-matagawa:
Done!

gulpfile.babel.js
```diff
@@ -81,7 +81,20 @@ gulp.task('core', function() {
...
+      'core/polyfills/MutationObserver*/MutationObserver.js',
+      'core/polyfills/childNodeRemove.js',
+      'core/polyfills/classList*/classList.js',
+      'core/polyfills/document-register-element*/document-register-element.max.js',
+      'core/polyfills/FastClick*/fastclick.js',
+      'core/polyfills/document-register-element*/innerHTML.js',
+      'core/polyfills/microevent.js*/microevent.js',
+      'core/polyfills/promise-polyfill*/promise.js',
+      'core/polyfills/setImmediate*/setImmediate.js',
+      'core/polyfill/viewport.js',
```
asial-matagawa: Oops, sorry... I will fix this

PR #1891, opened 2017-03-28
refactor(ons-tab): Sets 'active' attribute.

frandiox:
@asial-matagawa When a tabbar is reattached it shows again the original page instead of the last visible one due to this `active` attribute. That's troublesome in Vue bindings. Navigator or Splitter don't show the first page when they are reattached so I think tabbar should do the same.


asial-matagawa:
@frandiox 
Thank you!
I think at least the previous behavior is a bug.

Please feel free to merge with a new `CHANGELOG`.

PR #1892, opened 2017-03-28
Fix bugs on Safari related to asynchronous calls by replacing Custom Elements v1 polyfill.

asial-matagawa:
@frandiox (Cc: @anatoo @misterjunio @masahirotanaka )

This PR replaces Custom Elements v1 polyfill.

Old polyfill: [`document-register-element@1.0.7`](https://github.com/WebReflection/document-register-element/releases/tag/v1.0.7)
New polyfill: [`@webcomponents/custom-elements@1.0.0-rc.2`](https://github.com/webcomponents/custom-elements/releases/tag/v1.0.0-rc.2)

The new polyfill fixes 280 errors on Safari thanks to https://github.com/webcomponents/custom-elements/issues/34 .

Is it OK to merge this PR?

----

This change affects the browsers as follows:

## Safari (which has neither CE0 nor CE1)
**Before:** buggy
- `document.registerElement`:
    - `document-register-element`'s CE0 polyfill *(<- causes **delayed CE hooks**)*
- `customElements.define`:
    - `document-register-element`'s CE1 polyfill *(<- uses `document.registerElement` internally)*

**After:** fine
- `document.registerElement`:
    - `undefined`
- `customElements.define`:
    - `@webcomponents/custom-elements`'s CE1 polyfill *(<- works fine)*

## Chrome 1-32 (which has neither CE0 nor CE1)
Same as Safari.

## Chrome 33-53 (which has CE0)
**Before:** fine
- `document.registerElement`:
    - Native CE0 implementation *(<- used by the following CE1 polyfill)*
- `customElements.define`:
    - `document-register-element`'s CE1 polyfill *(<- uses `document.registerElement` internally)*

**After:** fine
- `document.registerElement`:
    - Native CE0 implementation *(<- not used at all)*
- `customElements.define`:
    - `@webcomponents/custom-elements`'s CE1 polyfill *(<- works fine)*

## Chrome 54- (which has CE0 implementation + CE1 implementation)
**Before:** fine
- `document.registerElement`:
    - Native CE0 implementation *(<- used by the following CE1 polyfill)*
- `customElements.define`:
    - ~Native CE1 implementation~ *(<- disabled because Onsen UI does not work with this)*
    - `document-register-element`'s CE1 polyfill

**After:** fine
- `document.registerElement`:
    - Native CE0 implementation *(<- not used at all)*
- `customElements.define`:
    - ~Native CE1 implementation~ *(<- disabled because Onsen UI does not work with this)*
    - `@webcomponents/custom-elements`'s CE1 polyfill

----
Only the problem of the new polyfill ([`@webcomponents/custom-elements@1.0.0-rc.2`](https://github.com/webcomponents/custom-elements/releases/tag/v1.0.0-rc.2)) is:

- It does not trigger `attributeChangedCallback` when `Element#className` is called.

So I have suppressed the unit test cases on `Element#className` temporarily.

----

This PR also includes:

- fix(unit-test): Suppress warnings about missing icon font files.
- fix(unit-test): Suppress warnings about HTTP subrequests on icon font files.
- chore(unit-test): Remove all `onlyChrome` from the test cases except for ones which are broken or cannot be run on Safari.
- chore(unit-test): Skip tests related to `Element#className` because the current polyfill does not support it.


frandiox:
@asial-matagawa That's great, thanks! Hopefully everything will be more consistent from now :sweat_smile: 
However, even though this PR is meant to fix issues, I think we should be a bit careful with this. Perhaps we can release the other master branch fixes in `2.2.1` and release this alone in `2.2.2` or even in `2.3.0`. What do you think?

asial-matagawa:
@frandiox 
I thought that too with the same reason.
I prefer `2.2.2` because `2.3.0` requires us to update the 4 bindings and the 12 Monaca templates.

I will postpone merging this.

----
I forgot to say:
- The current implementation depends on a bug of CE1 polyfill that we can read/write child elements in `constructor` (In the real CE1, we cannot do that and have to use Shadow DOM).
- In order to truly support CE1, we have to read/write child elements outside `constructor`.

Issue #1637, opened 2016-10-17
ons-toolbar come to follow in the ons-pull-hook in the iOS of safari


pinpikokun:
**Environment**
iOS 9, 10
safari

**Encountered problem**
ons-toolbar come to follow in the ons-pull-hook in the iOS of safari

**How to reproduce**
Please access the iPhone safari to the following url.
The list pull Then toolber will to follow.

**preview**
[link](https://pinpikokun.github.io/index2.html)

**source**
[link](https://github.com/pinpikokun/pinpikokun.github.io/blob/master/index2.html)

argelius:
@pinpikokun Thanks for reporting! We will take a look at this

@asial-matagawa Can you take a look at this iOS issue?

PR #1894, opened 2017-03-28
Release `onsenui@2.2.1`

asial-matagawa:
@frandiox @misterjunio @anatoo @masahirotanaka 

This is a PR for releasing `onsenui@2.2.1`.
Now all the bindings and the Monaca templates depend on `onsenui@2.2.x`, so I think we can release `onsenui@2.2.1` without any problems.

Note (about the current pull requests):
- #1885 has to be postponed to `2.2.2` or later due to the absence of @anataoo.
- #1892 should be released `2.2.2` or later because it has relatively big changes.
- #1887 is a fix for files in `bindings/react/`, so it need not be merged into this PR for now.

Is there any problem?
If not, I will update `CHANGELOG.md` and `package.json` and publish `onsenui@2.2.1`.


asial-matagawa:
Also, I will manually test Onsen UI with this branch

frandiox:
@asial-matagawa Thanks!

Does this contain everything from `master`? I guess we can directly merge `master` into `production` if that's the case.

asial-matagawa:
@frandiox 
Yes, this PR contains everything from `master`.
The reason why I created `release-2.2.1` branch is because if I use `master` branch, we cannot change `master` until it is merged into `production`.

I will use `master` directly in the next time.
Thank you for the comment!

asial-matagawa:
@frandiox 
The commit https://github.com/OnsenUI/OnsenUI/commit/b79ddbfc07105de985a99105ebb7db925279bb8c#diff-f96d6a24e6d6c2c79a80b239b7a738d5 on`ons-pull-hook.js` seems to have broken pull hooks on `Android 4.4.4 + Chrome 56` and `Android 5.0.1 + Chrome 56`.
For some reasons it worked on `Android 6.0 + Chrome 52`.

dsc_0863

Do you know the cause of this broken behavior?

asial-matagawa:
I am investigating the problem above now

asial-matagawa:
This problem seems a problem of the latest Android Chrome.
Even I used `onsenui@2.0.1` and `onsenui@2.2.0`, the same problem happened.

frandiox:
@asial-matagawa I see, then we don't need to check it right now. It doesn't happen in Cordova apps anyway.

asial-matagawa:
@frandiox 
I noticed your comment now.
I will postpone fixing this problem and fix it later with #1637.

PR #1910, opened 2017-03-31
fix(platform): ons.platform.isSafari() doesn't work in Safari 10

yarysh:
ons.platform.isSafari() doesn't work in Safari 10, this method returns `false` in it.

I've tested this solution in last Chrome, Firefox, Safari 10 and Safari Tech Preview.
My solution based on it: http://stackoverflow.com/a/9851769/1474569

### DIFF
--- a/core/src/ons/platform.js
+++ b/core/src/ons/platform.js
@@ -230,7 +230,7 @@ class Platform {
     if (this._renderPlatform) {
       return this._renderPlatform === 'safari';
     } else {
-      return (Object.prototype.toString.call(window.HTMLElement).indexOf('Constructor') > 0);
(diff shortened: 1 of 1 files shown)


frandiox:
@yarysh Thanks a lot for the PR!
@asial-matagawa @misterjunio Can any of you macboys have a look at this? :stuck_out_tongue:

asial-matagawa:
@frandiox 
I have tested some expressions on Safari 10.0.


This fix worked fine.
I would like to release this fix in the future release.

Thank you @yarysh !

misterjunio:
I got the same as @asial-matagawa

PR #1916, opened 2017-04-06
Splitter animations

frandiox:
@asial-matagawa @misterjunio We can finally deprecate `ons-sliding-menu` and `ons-split-view` :+1: 
If you want to have a look at the animations, use `animation="push"` or `animation="reveal"`.

It would be nice if you can test them a bit in Safari.


misterjunio:
@frandiox I tested the animations in Safari. The `reveal` seems totally fine, the `push` gets just a little sticky if I shake it really, really fast to both sides but not enough to worry I think. I would suggest to btw fix the core and angular1 examples of the splitter where the `splitter-side` on the right has a `threshold-ratio-should-open` attribute that I guess doesn't exist anymore and also update the documentation of the core and all bindings to reflect the new animations.

PR #1919, opened 2017-04-07
Release `onsenui@2.2.3`

asial-matagawa:


asial-matagawa:
For Onsen UI Team:
- I have added `core-js/fn/object/set-prototype-of` as a polyfill for `Object.setPrototypeOf` with global namespace pollution.

PR #1934, opened 2017-04-11
fix(angular1): Page loader fix.

misterjunio:
Page loader now throws 'destroy' event when page is unloaded.


asial-matagawa:
@misterjunio 
Thank you!
I will include this fix in the next release.

PR #1935, opened 2017-04-11
refactor(dialogs): All dialogs inherit from BaseDialogElement.

frandiox:
@asial-matagawa @anatoo This is using inheritance for dialogs and removes a bunch of duplicated functionality. It would be awesome if we could also "inherit" docs, although I guess there's nothing like that in the current wcdoc, right?


asial-matagawa:
@frandiox 
I agree with using inheritance for minimizing redundancy. Nice!
But I am not sure about things related to `wcdoc`.

@anatoo 
What do you think about the relation between this refactor and `wcdoc`?

frandiox:
@asial-matagawa Basically, we have long comments for `show`, `hide`, `onDeviceBackButton`, `visible`, `cancelable`, etc. repeated in 4 components (5 with action sheet). What I mean with "docs inheritance" is to provide a way to put all of these common docs in `BaseDialog` file and get them from there.

asial-matagawa:
@frandiox 
Yes, I know what you want to do.
What I am not sure is how to deal with the problem that `wcdoc` provides no way to do that.
(So I asked @anatoo)

PR #1937, opened 2017-04-12
ons.createElement

frandiox:
Not ready yet


frandiox:
@asial-matagawa Please have a quick look at this. It is trying to simplify the dialogs methods. Perhaps we can remove them if we implement `ons.createElement`.

PR #1938, opened 2017-04-12
fix(vue): Deriving events from the core.

frandiox:


asial-matagawa:
Note:
- This fix enables `vue-onsenui` to work with `onsenui@>=2.2.3` (which uses `@webcomponents/custom-elements`).
- After this fix, `vue-onsenui` cannot be used with `onsenui@<=2.2.2` (which uses `document-register-element`).

PR #1940, opened 2017-04-12
Return a promise from replacePage

airamrguez:
All public methods in the `Navigator` component return a promise except replacePage.


misterjunio:
@airamrguez thanks 👍

Issue #1340, opened 2016-04-05
Toolbar display not well while using both slide menu and navigator (Iphone only)


tienuit:
I use the example code (onsen 2 beta9)
http://codepen.io/frankdiox/pen/qEyvxB

but I got a problem on Iphone, the status bar and toolbar got something wrong
P1 - With both slide menu anf navigator
![img_2639](https://cloud.githubusercontent.com/assets/5275476/14279135/c3c235f2-fb55-11e5-9d03-8e73ded221ef.PNG)

P2 - Remove navigator
![img_2640](https://cloud.githubusercontent.com/assets/5275476/14279137/ca209eb6-fb55-11e5-8f07-dc7859a0476b.PNG)

IliaSky:
Thank you for the bug report and sorry for the late reaction. It should be fixed now.
Also if you run into similar problems and we don't respond in time you can disable the `auto status bar fill` feature and just manually add a `status-bar-fill` attribute to the element which you want to consume the iOS status bar.

PR #1944, opened 2017-04-12
fix iOS status-bar-fill not added in initially opened modals.

### LINKED ISSUE #1943: The status-bar-fill on iOS is not always added.
__Environment__

  
```
[Core]
  onsenui 2.2.4
  
[Framework]
  react 15.4.2

[Framework binding]
 react-onsenui latest

[Platform]
  Desktop - OS X 10.12.4 64-bit

[Browser]
  Desktop - Chrome 57.0.2987.133 on OS X 10.12.4 64-bit
```

__Encountered problem__

In the code below the modal is initially opened and the page inside of it does not have the `status-bar-fill` attribute.

simulator screen shot 12 apr 2017 12 32 21

__How to reproduce__
  Use the code below inside a Cordova project. Add the cordova-plugin-statusbar plugin. Run it on iOS.

  
  [__Demo link__](https://tutorial.onsen.io/?issue)

  

- __HTML__

```html
<div id="app"></div>
```

- __JS__

```jsx
import cordova from 'cordova';
import PropTypes from 'prop-types';
import React, { Component } from 'react';
import { render } from 'react-dom';

import 'onsenui';
import { Page, Button, Splitter, SplitterContent, Modal, Toolbar } from 'react-onsenui';

import './css/index.css';

class App extends Component {
  renderToolbar = () => {
    return (
      <Toolbar>
        <div className="center">
          Modal initially opened
        </div>
      </Toolbar>
    );
  }
  renderModal = () => {
    return (
      <Modal isOpen={true}>
        <Page renderToolbar={this.renderToolbar}>

        </Page>
      </Modal>
    );
  };
  render() {
    return (
      <Page key="tey">
        <Splitter>
          <SplitterContent
            SplitterSide
            style={{
                boxShadow: '0 10px 20px rgba(0,0,0,0.19), 0 6px 6px rgba(0,0,0,0.23)'
            }}
            side='left'
            width={200}
            collapse={true}
            isSwipeable={true}
          >
            <Page renderModal={this.renderModal}>

            </Page>
          </SplitterContent>
        </Splitter>
      </Page>
    );
  }
}

function onDeviceReady() {
  document.removeEventListener('deviceready', onDeviceReady);
  StatusBar.styleDefault();
  render(
    <App isDevice={cordova.platformId !== 'browser'} />,
    document.getElementById('root')
  );
}

document.addEventListener('deviceready', onDeviceReady, false);

```

airamrguez:


anatoo:
@airamrguez Sorry for the delay. Thanks!

@frandiox @misterjunio @asial-matagawa This changes seems to be fine. I will merge this changes to master branch. Let me know if there is a problem.

frandiox:
@airamrguez @anatoo I think this PR is making `status-bar-fill` attribute to be present in more than 1 page, which is wrong for nested pages.

`findParent` only checks the first parent and returns `null` if it's not a modal (most of the times). It doesn't check further ancestors to see if `status-bar-fill` is there.

Can we remove this?

airamrguez:
@frandiox I think I have a fix for this. I’ll try to submit a pr for this today.

frandiox:
@airamrguez Thanks. Just removing the negation `!` or the whole thing works well, but I'm not sure what's necessary to keep the modal fix.

airamrguez:
@frandiox To keep the modal fix I end up commenting the `_hasAPageControlChild` from `_tryToFillStatusBar()` and changing the query from the `findParent` call:

```js
_tryToFillStatusBar(){
    internal.autoStatusBarFill(() => {
      const filled = util.findParent(this, e.hasAttribute('status-bar-fill') && e.parentNode && e.parentNode.nodeName.match(/ons-(splitter|sliding-menu|navigator|tabbar)/i));
      util.toggleAttribute(
          this,
          'status-bar-fill',
          !filled && (
                    this._canAnimateToolbar() /* ||
                    remove this check -> !this._hasAPageControlChild() */)
          );
    });
  }
```
I've been using this patch in an application that has an initially opened modal and nested pages since a while and it's working, but I'm not sure if I'm missing any other component name in the regex. What do you think about these changes?

In the `findParent` function the `until` argument can be removed because it was introduced with this pull request.

frandiox:
@airamrguez `hasAPageControlChild` is there to fix another situation (#1340), and that regex would get too complex because pages can be inside divs, body or dynamically added wrappers.

Can you confirm that this is the situation you want to keep? https://codepen.io/frankdiox/pen/jGBZay?editors=0010

That works in 2.6.0 and still works when I remove `, e => !e.nodeName.match(/ons-modal/i)` for the code. I guess it should be fine.

airamrguez:
@frandiox yes, this is the situation. I'm going to install 2.6.0. Let me check this out.

airamrguez:
@frandiox I've made some changes to the pen you posted to mimic my component hierarchy. The initially opened modal is working ok with the latest version, but there is another problem with nested pages. There is a gap between nested pages when the inner page does not have a toolbar.

[https://codepen.io/airamrguez/pen/NapeBV](https://codepen.io/airamrguez/pen/NapeBV)

frandiox:
@airamrguez Yes, that's one of the things I was fixing but then I got the problem with the modal thing and I came to this PR :sweat_smile: 

I'll check that code tomorrow and try to find a solution :+1:

airamrguez:
@frandiox 🙏🏽 thanks for your time and dedication

frandiox:
@airamrguez Fixed in https://github.com/OnsenUI/OnsenUI/pull/2194/commits/0b4006a6821c273c331adcdb9e8e8183e10afe2b#diff-52c49156be6687c4b5dfefe178df4972R187

The child pages were filling the status bar before the parent page. I tried your codepen with this version and everything works :+1:

airamrguez:
@frandiox Cool!! looking forward to test this

airamrguez:
@frandiox I'm sorry but I've found another use case in which the iOS status bar does not work ok with the latest version of OnsenUI.
[https://codepen.io/airamrguez/pen/NapeBV?editors=1010](https://codepen.io/airamrguez/pen/NapeBV?editors=1010)

In the demo, if you remove the comment around the splitter then it will work.

frandiox:
@airamrguez My bad, I added another fix after that which prevents modal pages from getting `status-bar-fill`. However, I think this is now the correct behavior for the core.

The problem after this is the way React bindings create `ons-modal`. It is placed in a `.page__extra` div inside the page itself instead of using a "portal" like all the other dialogs. I've changed this in my local version and now the modal can be used normally without `renderModal` function. The only problem is that `onShow` and `onHide` would be renamed to `onPostShow` and `onPostHide` like all the other dialogs (and also `onPreShow` and `onPreHide` would be available). What do you think about that change for `react-onsenui@1.7.0`? --- Commit here: 67b5c611e6f8a2d4d27b3c6c24823b3d1c2350dd

airamrguez:
@frandiox adding the mockStatusBar definitely saves a lot of time. That was huge!

I think making the Modal component share the same API that BaseDialog makes the API more consistent with the rest of dialog components.

May be in the 1.7.0 version can use React 16 and replace ReactDOM.unstable_renderSubtreeIntoContainer with ReactDOM.createPortal.

Upgrade should be at no cost as it's said in the docs:

> With minor exceptions, if your app runs in 15.6 without any warnings, it should work in 16.

frandiox:
@airamrguez Yes, I've been checking the new portal API and will replace the old one when I get some time :+1:

frandiox:
@airamrguez I've implemented Dialogs with `ReactDOM.createPortal` new API, but I've found a problem. A dialog could be unmounted while showing (pop a page that contains a toast, for example), which would make it disappear without any animation. Currently it's possible to delay the dialog unmount with code [like this](https://github.com/OnsenUI/OnsenUI/blob/master/bindings/react/src/components/BaseDialog.jsx#L77), but I don't know any alternative when using `createPortal`. Do you have any idea?

airamrguez:
@frandiox I don't know if I have misunderstood the problem. I've tried some different scenarios here [https://codepen.io/airamrguez/pen/Nagwdb?editors=1010](https://codepen.io/airamrguez/pen/Nagwdb?editors=1010)

I think the animations are being completed before the parent components are unmounted. Otherwise, can you share some use case on a CodePen?

frandiox:
@airamrguez Thanks for the code! I'll try to come back to this later this week when I have some time, right now I need to focus on something else. Basically I think this is the same issue https://github.com/facebook/react/issues/10826
When a page is unmounted (parent component), the dialog (child component) will also be unmounted synchronously even if its DOM node is not rendered inside the parent. I guess this could be tricked in some way, like rendering `null` and just creating/destroying the dialog asynchronously.

frandiox:
@airamrguez I think the situation I described is kind of similar to what happens in your codepen when clicking on "Pop page from modal". Modal starts fading out but at some point the page animation finishes and removes the modal element. As a result, the modal animation ends abruptly.

I think we'll need to stick to `unstable_renderIntoContainer` for now :/

Issue #1952, opened 2017-04-18
Core | Tabbar issue: Carousel fires unexpected postchange


misterjunio:
__Environment__

```
[Core]
  onsenui latest
  
[Platform]
  Desktop - OS X 10.12.4 64-bit

[Browser]
  Desktop - Chrome 57.0.2987.133 on OS X 10.12.4 64-bit
```

__Encountered problem__
Having an `ons-carousel` inside an `ons-tabbar` fires an unexpected `postchange` event when the window is resized, even if another tab is active.

__How to reproduce__
Link below. Here I'm using a video loading but any trigger to resize, for example also changing the orientation of a device (or simulator) launches `postchange` on the carousel.

[__Demo link__](https://tutorial.onsen.io/?issue)

- __HTML__

```html
<ons-tabbar ng-controller="MyController">
    <ons-tab label="Tab 1" page="tab1.html" active>
    </ons-tab>
    <ons-tab label="Tab 2" page="tab2.html">
    </ons-tab>
  </ons-tabbar>

  <ons-template id="tab1.html">
    <ons-page>
      <ons-toolbar>
        <div class="center">Tab 1</div>
      </ons-toolbar>

      <ons-carousel fullscreen swipeable overscrollable auto-scroll>
        <ons-carousel-item style="background-color: gray;">
          <div class="item-label">GRAY</div>
        </ons-carousel-item>
        <ons-carousel-item style="background-color: #085078;">
          <div class="item-label">BLUE</div>
        </ons-carousel-item>
        <ons-carousel-item style="background-color: #373B44;">
          <div class="item-label">DARK</div>
        </ons-carousel-item>
        <ons-carousel-item style="background-color: #D38312;">
          <div class="item-label">ORANGE</div>
        </ons-carousel-item>
      </ons-carousel>
    </ons-page>
  </ons-template>

  <ons-template id="tab2.html">
    <ons-page>
      <ons-toolbar>
        <div class="center">Tab 2</div>
      </ons-toolbar>

      <video controls>
      </video>
      <a href="#" onClick="javascript:vidSwap('http://techslides.com/demos/sample-videos/small.mp4'); return false;">Link</a>

    </ons-page>
  </ons-template>
```

- __JS__

```javascript
document.addEventListener('postchange', function (e) {
  console.log('event postchange', e.target);
});

function vidSwap(vidURL) {
  var myVideo = document.getElementsByTagName('video')[0];
  myVideo.src = vidURL;
[...]

PR #1954, opened 2017-04-18
Carousel fix

misterjunio:
@frandiox @asial-matagawa Here is a fix for issue #1952. `postchange` event listener is added and removed on show and hide of the `ons-carousel`.


frandiox:
Thanks!

Issue #1928, opened 2017-04-09
Style of ons-list-header is broken


puku0x:
__Environment__ 

```
[Core]
    onsenui 2.2.3
```

__Encountered problem__

I found the style of `ons-list-header` has been broken since v2.2.0.

- No `border-bottom`
- No `padding-top`
- Wrong `background-color`

>Onsen UI ~2.0.5
><img width="300px" src="https://cloud.githubusercontent.com/assets/2607019/24838738/0d739f78-1d89-11e7-8c47-74b3d04af566.png"/>

>Onsen UI 2.2.0
><img width="300px" src="https://cloud.githubusercontent.com/assets/2607019/24838753/51f5327e-1d89-11e7-83a6-3305b995a5df.png"/>

asial-matagawa:
@puku0x 
Thank you for reporting!

@anatoo 
Is this a bug?
Or an expected behavior?

anatoo:
@puku0x @asial-matagawa This is expected appearance. `ons-list-header` is originally implemented as a list category header.
In 2.0.x, `ons-list-header` is suddenly changed by other developers, so I just restored to original appearance in 2.2.0.

puku0x:
@anatoo Thank you for your response.

However, I think the previous one also looks good because it is similar to iOS's list header 🙂 

If you have time, could you consider providing both styles of header ?
(Like Ionic's `ion-list-header` and `ion-item-divider`, for instance)

anatoo:
Sorry for the delayed response because I am part-time developer for Onsen UI.

> If you have time, could you consider providing both styles of header ? (Like Ionic's ion-list-header and ion-item-divider, for instance)

Ok, it's should be fine. I will propose new components for that to Onsen UI team.

puku0x:
@anatoo Thank you very much !

PR #1960, opened 2017-04-19
Added ons-list-title element.

anatoo:
@frandiox @asial-matagawa @misterjunio  I added `ons-list-title` elements.

We have already `ons-list-header` elements, but this elements represent list category header or list divider not list title label. So I added `ons-list-title` elements for that.

list_demo___onsen_ui

Could you review this changes?

Related issue: https://github.com/OnsenUI/OnsenUI/issues/1928


frandiox:
@anatoo Is this supposed to be used outside `ons-list`? Can it be included inside as well?

anatoo:
@frandiox Yes, it should be outside `ons-list`.

PR #1962, opened 2017-04-19
Fix sourcemap of `onsenui.js` by replacing Rollup with webpack 2.

### LINKED ISSUE #1958: Use webpack 2 to fix sourcemap of `onsenui.js`
@frandiox @misterjunio @anatoo 

I would like to replace Rollup with webpack 2 in order to fix the sourcemap of `onsenui.js`.
(For example, in `onsenui@2.2.4`, we cannot set a breakpoint on L153, L156 and L161 in Chrome DevTools as illusrated below)

What do you think?

### Detail

In onsenui@2.2.4, when we set a break point on some point of `onsenui.js` in Chrome DevTools, sometimes we are navigated to a wrong line of original file.

I found that this is caused by [a change](https://github.com/OnsenUI/OnsenUI/compare/2.0.3...2.0.4) committed between `onsenui@2.0.3` and `onsenui@2.0.4`.
This change replaces `rollup@0.26.7` with `rollup@0.36.3`, but `rollup` made [a breaking change](https://github.com/rollup/rollup/blob/master/CHANGELOG.md#0323) in `rollup@0.32.3` which makes sourcemaps incompatible with Chrome DevTools.

Before `rollup@0.32.3` (first letter of each line is mapped correctly):

After `rollup@0.32.3` (first letter of each line is not mapped):

We can roll back `rollup` to an old version, but I think using webpack 2 is more decent decision.

This problem (= weird(?) behavior of Chrome DevTools) is recently discussed also in the official repos of [`rollup`](https://github.com/rollup/rollup/issues/1370) and [`webpack`](https://github.com/webpack/webpack/issues/2145#issuecomment-294361203).

asial-matagawa:
Closes #1958.


asial-matagawa:
I'm stuck by the errors on CircleCI above...

The exactly same JavaScript code worked on my environment (macOS Sierra + Chrome 58) but did not work on Circle CI (Linux + Chrome 58).
I guess ChromeDriver for Linux or Chrome for Linux is broken but it's very weird...
What I did is only to replace Rollup with webpack 2.

asial-matagawa:
@anatoo @misterjunio @frandiox 
If you have a Linux environment, could you run `gulp test` in `bindings/angular2`?
CircleCI (= Linux instance) reported the following error but it was not reproduced in my environment.

```
[05:12:12] Starting 'e2e-test'...
[05:12:14] I/direct - Using ChromeDriver directly...
[05:12:14] I/launcher - Running 1 instances of WebDriver
Started
.......................................F...........[05:13:44] W/element - more than one element found for locator By(css selector, ons-tab) - the first result will be used
..[05:13:46] W/element - more than one element found for locator By(css selector, .normal-page) - the first result will be used
[05:13:46] W/element - more than one element found for locator By(css selector, .normal-page) - the first result will be used
..

Failures:
1) pull-hook.html should load stuff when pulled down
  Message:
    Failed: Wait timed out after 5020ms
```

asial-matagawa:
This bug was reproduced on my environment (Debian + Chrome 58).

asial-matagawa:
This issue has been fixed and it was due to:
- Change detection was not run event after new `ons-list-item` was pushed
- But the E2E test codes managed to run change detection only on macOS for unknown reason.
    - I guess it's related to the way that Zone.js works.

PR #1963, opened 2017-04-20
Action sheet

frandiox:
@misterjunio @asial-matagawa Please have a look at this and test it on Safari. For the "notification-like" method, what do you think it's better? `ons.notification.action(...)`? `ons.actionSheet(...)`?


misterjunio:
@frandiox looks good on Safari 👍  and I think `ons.actionSheet()` would be more intuitive

asial-matagawa:
@misterjunio 
Thank you for testing, but actually iOS Safari behaves differently compared to macOS Safari.
Could you test also on iOS devices if you have time?

asial-matagawa:
@misterjunio 
For example #1870 occured on iOS 9, 10

misterjunio:
@asial-matagawa sure, will do

misterjunio:
@asial-matagawa tested and working in all iOS versions
@frandiox I suppose the font thing is some issue in my environment. Other than that maybe you can change the example title to `Action sheet` and make the buttons hide it

asial-matagawa:
@misterjunio 
Thank you!
I am relieved to hear that.

frandiox:
I remembered about #1870 when making the animations, should be fine :sweat_smile:

frandiox:
I added `ons.actionSheet({ buttons: [...], title: '...'});` method. I guess there's no problem with this either.

frandiox:
@misterjunio @asial-matagawa I think this is done. It includes all the bindings except Angular2. I uploaded examples in case you want to check.

misterjunio:
Everything looks good on Safari in desktop and iOS 10 👍

anatoo:
@frandiox 
I have a opinion about the method name.

I think we should prefer `ons.openActionSheet()` or `ons.createActionSheet()` not `ons.actionSheet()` because `actionSheet` isn't a verb unlike `confirm` and `alert` and `prompt`.

frandiox:
@anatoo I think `ons.createActionSheet` could be confusing because we are using `ons.createDialog`, `ons.createAlertDialog` and `ons.createPopover` for something different. `ons.openActionSheet` could be fine but I guess simply `ons.actionSheet` is still understandable. Maybe we should go with something different like `ons.requestAction` or `ons.notification.action`?

@asial-matagawa @misterjunio What do you think?

anatoo:
@frandiox 

>I think ons.createActionSheet could be confusing because we are using ons.createDialog, ons.createAlertDialog and ons.createPopover for something different. 

Certainly. Then I prefer to choose `ons.openActionSheet()`. `ons.actionSheet()` seems to be inconsistent because we are already using verb for naming method like `createDialog()`.

misterjunio:
Hum I think @anatoo has a point with the verb thing and `ons.notification.action` would have the same problem but you're also right that `ons.create-` gives the wrong idea. I would go for either `ons.openActionSheet` or `ons.requestAction`

asial-matagawa:
@frandiox @anatoo 
I think `ons.actionSheet()` is great but it would be better to have a certain prefix for it because we will have more components similar to action sheets.

If the methods have a common prefix, we will have:
```
ons.openActionSheet()
ons.openToast()
ons.openDatePicker()
```
but if not, we will have:
```
ons.actionSheet()
ons.toast()
ons.datePicker()
```
The latter naming could cause a problem when we sort the methods of `ons` alphabetically.

----
If we have `ons.notification.action`, we will suffer from how to deal with components like date pickers, so I think we should avoid using `ons.notification.action`.

frandiox:
So the name has been decided: `ons.openActionSheet` :+1:

asial-matagawa:
Thank you for discussing this!

PR #1965, opened 2017-04-21
Card

misterjunio:
@frandiox @asial-matagawa this is the PR for the new `ons-card`. I will wait for @anatoo to implement the `ons-card-title` and `ons-card-content` to finish the component and create a nicer example.


misterjunio:
@frandiox @asial-matagawa I think the `ons-card` is ready. Please review and merge it when you find time 👍

asial-matagawa:
@misterjunio 
I checked the implementation and I think it's fine.

----
We can also compile
```html
<ons-card>Lorem ipsum ...</ons-card>
```
into
```html
<ons-card>
  <div class="card__content">
    Lorem ipsum ...
  </div>
</ons-card>
```
as `ons-page` is doing so, but cards have no "main" block unlike pages as long as I read [the Material Design guidelines for Cards](https://material.io/guidelines/components/cards.html#cards-content-blocks).
Moreover, we might support `Rich media`, `Actions` and some blocks described in the guideline.

Therefore I think the current specification which forces users to use the following structure is fine for now.
```html
<ons-card>
  <div class="title">...</div>
  <div class="content">...</div>
</ons-card>

```

asial-matagawa:
@misterjunio 
Apart from the specification of slots for `ons-card`, it would be better if the current example (screenshot is below) is more similar to real use cases. For example, I guess buttons should be flat ones.



"real use cases" means like this: https://storage.googleapis.com/material-design/publish/material_v_11/assets/0Bzhp5Z4wHba3bG0tNkIyNDhJZEU/components_cards28.png

This is something optional, so please feel free to concentrate on other things.

misterjunio:
@asial-matagawa about the format,
```html
<ons-card>
  <div class="title">...</div>
  <div class="content">...</div>
</ons-card>
```
is merely a possibility, obviously users can add any other type of content they wish.

About the example I tried to tweak it a little bit, not sure if it's closer to what you meant.

Do you think that in general we should provide more 'building blocks' for the cards? Like specific action buttons and things like that? If so I'll need to ask @anatoo for the CSS..

asial-matagawa:
@misterjunio 
>merely a possibility
Ah, I see.

>Do you think that in general we should provide more 'building blocks' for the cards?
I think we can postpone supporting more building blocks since what we need to do for now is only to give enough extendability to `ons-card` for future changes.
Let's release `onsenui@2.3.0` with `ons-card` which contains the current minimum features.

Any suggestions?

PR #1969, opened 2017-04-24
Release `onsenui@2.2.5`

asial-matagawa:
### Postponed to `2.3.0`

- 9f5e02c32a1a8444b1a158e0ff651e88dcc5c9fb


asial-matagawa:
64d9ed1dc17c074fd35d2bc144decc6e9e36877f seems to break Espree during `gulp build-docs`.
I will check the cause of this problem

frandiox:
@asial-matagawa Oh, probably there is an error in the docs comments. I saw this but actually forgot about it.

asial-matagawa:
@frandiox 
I have identified the cause of this problem.
You used `traling commas in function argument list` in `RevealSplitterAnimator`, but `wcdoc@0.1.18` cannot recognize ES2017 things because it has [a wrong setting](https://github.com/OnsenUI/wcdoc/blob/da172cac98de5d79c16f19dada75fcf0d223dc88/src/ParsedFile.js#L86).
We have to fix [`OnsenUI/wcdoc`](https://github.com/OnsenUI/wcdoc)...

frandiox:
Oh, I see, good catch :)

Issue #1976, opened 2017-04-25
Deprecate `*Factory` in `angular2-onsenui`.


asial-matagawa:
@frandiox (Cc: @misterjunio @anatoo)

I thought about the necessity of `*Factory` in `angular2-onsenui`.

Pros of `*Factory`
  - We can avoid polluting component templates with dialog definitions
  - We can pass parameters to dialog components as if we pass parameters to `ons-page`s with `pushPage`
  - We can provide ways which are equivalent to `ons.create*` in the core

Cons of `*Factory`
  - We have to inject `*Factory` to the component where we want to call `*Factory.create*(~~~~)`
  - We have to destroy dynamically created components in `ngOnDestroy` hook
  - We have to register all dialog component classes to `declarations` and `entryComponents` in NgModule
  - We have to use host bindings in order to attach @Input/@Output to dynamically created components
  - Dynamic component creation in JS frameworks essentially tends to cause complexity and should not be exposed to users

Conclusion
  - We should deprecate `*Factory` and use only in-template dialog definitions as the other bindings are doing so

Even without `*Factory`, we can write code as follows:
```ts
import {
  Component,
  ComponentRef,
  ViewChild,
  Params,
  AfterViewInit,
  OnDestroy,
  OnsenModule,
  NgModule,
  onsNotification,
  CUSTOM_ELEMENTS_SCHEMA
} from 'angular2-onsenui';
import {platformBrowserDynamic} from '@angular/platform-browser-dynamic';

@Component({
  selector: 'app',
  template: `
  <ons-page class="page">
    <ons-toolbar>
      <div class="center">Alert Dialog</div>
    </ons-toolbar>
    <div class="content">
      <ons-list>
        <ons-list-header>Custom dialogs</ons-list-header>
        <ons-list-item tappable (click)="alert.show()">Custom alert-dialog</ons-list-item>
        <ons-list-item tappable (click)="dialog.show()">Custom dialog</ons-list-item>

        <ons-list-header>Notification</ons-list-header>
        <ons-list-item tappable (click)="alert()" id="alert-button">Alert</ons-list-item>
        <ons-list-item tappable (click)="confirm()" id="confirm-button">Confirmation</ons-list-item>
        <ons-list-item tappable (click)="prompt()" id="prompt-button">Prompt</ons-list-item>
      </ons-list>
    </div>

    <ons-alert-dialog cancelable #alert>
      <div class="alert-dialog-title">Warning!</div>
[...]

PR #1982, opened 2017-04-27
Toast

frandiox:
@OnsenUI/members I've implemented a proposal for `ons-toast` here. Let's discuss what to change or how to implement the rest of the features. Animations are already finished (ascend, lift, fall, fade, none).

`ons-toast` is a dialog without mask that inherits from `base-dialog` and implements all its methods/features (show, hide, toggle, deviceBackButton, etc).
The current API is:
```html
1: <ons-toast>Description</ons-toast>
2: <ons-toast>Description<button>Action</button></ons-toast>
3: <ons-toast><div class="message">Description</div><div class="button">Action</div></ons-toast>
```
Or any combination of those 3. Do you think we should allow `<button>`, `class="button"`, or both? Raw text/tags, `class="message"`, or both?

Apart from that, new features to be implemented:

1. Timeout: Allow `show({timeout: 3000})` and call `hide` automatically after 3 seconds.
2. `ons.notification.toast` (opinions about the method name?): Allows providing a message, button label, callback, timeout, etc.
3. Create a "Toast Controller" that stores a queue of toasts and displays them 1 by 1 (by the MD specs, no more than 1 toast should be displayed at a time).
4. Preset modifiers: `success`, `warning`, `error`, `info`, etc. (changing background and text color). @anatoo What do you think about providing these modifiers?

Since the basics are already implemented, feel free to test them (especially on Safari and iOS). These animations are a bit different so #1870 could still be a problem.

Any other feedback is also welcome :+1:


misterjunio:
Looks great.
About the API I think there is no problem in leaving all the options open, some users may prefer one or another.
1. Actually I think the preferred behavior is that the toast fades out automatically but to make it default we'd have to tweak the API a little bit I guess so maybe it's fine;
2. `ons.notification.toast` sounds very natural;
3. I can't imagine a time when the user wants to receive more than one toast at once so I wouldn't do this;
4. Don't know if that feels native as I don't think I ever saw that.
As an extra I would say that my idea of a toast (may be wrong) is that it's a kind of alert that targets more the app in general instead of a specific page, you know? So perhaps we could find a way to keep it alive?

anatoo:
@frandiox 

> Preset modifiers: success, warning, error, info, etc. (changing background and text color). @anatoo What do you think about providing these modifiers?

It should be OK to have. But they are not absolutely necessary at now. I prefer to have no preset modifiers at the start.

frandiox:
Updates:

1. Sometimes a timeout won't be necessary when buttons like "Undo" or "Accept" are provided, so I think it is easier to manually provide a timeout only when needed than overwriting an existing behavior. It is also hard to find a proper default timeout that works for most cases (4, 5 seconds?). Therefore, I'd prefer to have no timeout by default. Also, I think the timeout can be implemented only for `ons.notification.toast` method, not for the element's `show`.
2. `ons.notification.toast` implemented.
3. Regarding @misterjunio 's input, I think notifications can come from the server at a "random" timing so perhaps we can provide this option in order to queue the notifications and follow the specs.
4. Let's skip this then, but I'd like to write a blog post or something about how to create simple modifiers for `ons-toast`.

If there's no other feedback I'll proceed in this way :+1:

asial-matagawa:
@frandiox 
I checked the guideline of snackbars and I have no new opinions about 1~4.
Great work!

I would appreciate if you merge this PR after everything is done.

asial-matagawa:
@anatoo @frandiox 
Material toast message looks a bit large compared to the image in #1908.


It seems that this is caused by a fix (cd7d975cd019e9f628189bb8b5f966e7b0d7a118).
Is this a correct behavior?

frandiox:
@asial-matagawa Good catch.
After fixing that CSS issue I think this is done. I implemented components for React and Vue, not for Angular2 since I guess we are deprecating factories (#1976). However, `onsNotification.toast(...)` should still work in Angular2.

asial-matagawa:
@frandiox 
This CSS issue was reported in #1991 and fixed by #1999.
Now it should be fine

asial-matagawa:
Sorry, it's related #1991 but did not fixed.

asial-matagawa:
@anatoo 
`--material-font` still has a problem even after #1999 is merged.
Could you check this?

anatoo:
@asial-matagawa What's the problem? Please tell me more.

asial-matagawa:
@anatoo
Sorry, what I meant is https://github.com/OnsenUI/OnsenUI/pull/1982#issuecomment-300097883.
Is this large-message problem already fixed?

anatoo:
@asial-matagawa I fixed on #1999.

asial-matagawa:
@anatoo
Thank you!
I will test it on real devices before release `onsenui@2.3.0`.

PR #1983, opened 2017-04-27
fix(core): Remove for...of loop.

frandiox:
@asial-matagawa Please have a look at this. I think it should be enough.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ dev
  * ons-input: Fixed [#1974](https://github.com/OnsenUI/OnsenUI/issues/1974).
  * ons-tabbar: Minor fix for situations where the '.page__content' is provided.
  * ons-splitter: Fixed timing for initial animations in bindings.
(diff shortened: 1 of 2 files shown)


asial-matagawa:
@frandiox 
Thank you!

FYI: Actually [`ons-select.js` also has `for...of`](https://github.com/OnsenUI/OnsenUI/blob/ac6bf2ab0c238a3f036dd2ad2322284b53b6f5a6/core/src/elements/ons-select.js#L230) but it's safe because if the expression after `of` is an array literal it is compiled into a code without `Symbol`.

frandiox:
@asial-matagawa Only with literals? That's interesting...

asial-matagawa:
@frandiox 
This will help: https://www.npmjs.com/package/babel-plugin-transform-es2015-for-of#optimization

frandiox:
@asial-matagawa Yep, that's the same code, I guess we can merge this and release tomorrow :)

asial-matagawa:
♨️

PR #1993, opened 2017-05-06
Allow change the defaultPageloader

badlee:
Now for change the default page loader, we can write :
```js
ons.defaultPageLoader(new ons.PageLoader(myLoader, myUnLoader));
```

instead of :

```js
ons.defaultPageLoader._loader = myLoader;
ons.defaultPageLoader._unloader = myUnLoader;
```

### DIFF
--- a/core/src/ons/ons.js
+++ b/core/src/ons/ons.js
@@ -55,7 +55,9 @@ ons._animationOptionsParser = animationOptionsParser;
 ons._autoStyle = autoStyle;
 ons._DoorLock = DoorLock;
 ons._contentReady = contentReady;
-ons.defaultPageLoader = defaultPageLoader;
+ons.defaultPageLoader = (pageLoader)=>{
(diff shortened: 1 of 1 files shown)


core/src/ons/ons.js
```diff
@@ -55,7 +55,9 @@ ons._animationOptionsParser = animationOptionsParser;
 ons._autoStyle = autoStyle;
 ons._DoorLock = DoorLock;
 ons._contentReady = contentReady;
-ons.defaultPageLoader = defaultPageLoader;
+ons.defaultPageLoader = (pageLoader)=>{
+   return defaultPageLoader=pageLoader;
```
misterjunio: @badlee it's a nice idea but shouldn't the default one be kept if the user doesn't provide a custom one?

Like `return pageLoader ? defaultPageLoader = pageLoader : defaultPageLoader;`

Issue #1831, opened 2017-02-21
Deprecate features/components


frandiox:
@anatoo @asial-matagawa @misterjunio 

We have talked about deprecating some stuff already and I just want to make a list so we don't forget for the next release:

- [x] Theme Roller.
- [x] Stylus-lang.
- [x] `ons-template`: use `template` instead.
- [x] `ons-sliding-menu` (Angular1): we have `ons-splitter`. -- _Added warning_.
- [x] `ons-split-view` (Angular1): we have `ons-splitter`. -- _Added warning_.
- [x] `loadPage` method of `ons-tabbar`.
- [x] `options.refresh` for `popPage`: Do you think we need this in Onsen UI v2? I think `ons-init` and `ons-show` handlers could be enough in Angular1 to refresh the previous page. -- _Added warning_.

Feel free to add or discuss anything else.

anatoo:
👍

sdmgeek:
How to refresh previous page when clicking ons-back-button if "options.refresh for popPage" is deprecated? Just using JavaScript Core v2.2.6.

frandiox:
@sdmgeek What is your use case? Do you just want to run the previous page's controller (Angular controller) again? I think that can be done in a page lifecycle event such as `show`.

sdmgeek:
Thanks for reply~
I'm not using Angular, just using JavaScript.
The use case is that the child page call the server API that may affect its parent page, so I want to refetch from the parent page url to get whole fresh content when clicking the back button.
Is there other way to achieve the refresh?

frandiox:
@sdmgeek `options.refresh` is just removing the parent page and creating it again. I guess you are calling some function on `init` event, so you just need to call the same function on `show`, `hide`, `prepush` events or override the `ons-back-button` behavior (with `onClick` dom property) in order to pass a callback. I think it would be enough with any of these possibilities.

sdmgeek:
Got it.
Is it suggested to use `navigator._pageLoader.unload` to unload the parent page and `insertPage` to insert the page again? 
In addition, are there any ways to check the content of page going to be pushed at the `prepush` event callback?

frandiox:
@sdmgeek `pageElement._destroy(); pageElement.remove();` is probably enough. You can do that and then insert it again but the idea is that you update the parent page before popping. I think you mean `prepop` event. You can simply get the navigator and access to the previous page with `navigator.pages[navigator.pages.length - 2]` or something similar.

[...]

PR #2003, opened 2017-05-12
chore(ons-navigator): Remove `options.refresh`.

asial-matagawa:
Related issue: #1831

@frandiox 
This PR removes `options.refresh` for `ons-navigator#popPage`.
Is it safe to remove?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ dev
  * core: Added new method `ons.createElement(...)` that allows creating new elements from templates or inline HTML ([#1941](https://github.com/OnsenUI/OnsenUI/issues/1941)).
  * core: Added `ons-card` element.
  * core: A fake device back button event is now fired on ESC press.
+ * ons-navigator: Added `removePage` method.
  * angular1: Added `ons-action-sheet` bindings.
  * angular1: Added `ons-card` bindings.
  * angular1: Added `ons-list-title` bindings.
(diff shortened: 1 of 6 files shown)


frandiox:
@asial-matagawa Thanks for the PR!
A user suggested adding `navigator.removePage` if we remove `options.refresh`, what I think it could be fine. What do you think about that? It simply calls `pageLoader.unload(pageElement)` or `pageElement._delete()`.

asial-matagawa:
@frandiox 
I think it would be fine if `navigator.removePage` is provided since it is easy for users to find out from onsen.io.
But we have to release 2.3.0 soon, let's postpone it to 2.3.1 and later if we provide `navigator.removePage`.

frandiox:
@asial-matagawa I mean that, if we deprecated `options.refresh`, perhaps we should provide `removePage` in the same release so it is easier for some users to update. I think it is a very easy to implement method. I don't mind if we postpone it but perhaps we should also postpone this PR then?

asial-matagawa:
@frandiox 
Ah, I see.

>I think it is a very easy to implement method.
Actually I don't have enough knowledge to make changes to the core, so I could not estimate the cost. Could you implement `removePage` and related things? I am not sure if we need to make changes to the bindings as well.

asial-matagawa:
@frandiox
I have implemented `removePage` for both of `core` and `angular-onsenui`.
I will merge this.

frandiox:
@asial-matagawa Thanks!

PR #2007, opened 2017-05-16
chore(onsenui.js): Bundle polyfills with `import` clause instead of concatenating them.

asial-matagawa:


asial-matagawa:
In this PR, the order of loading libraries has changed in the commit 9c0d0b03e819b6801d26393e4881f3520b1823e9 as follows:

**Before:**
```js
MutationObserver polyfill
FastClick
microevent.js
setImmediate polyfill
viewport.js
```

**After:**
```js
MutationObserver polyfill
setImmediate polyfill
FastClick
microevent.js
viewport.js
```

But at least FastClick, microevent.js and microevent.js don't depend on each other, I believe that this change caused no issues.

PR #2005, opened 2017-05-15
chore(core): Removed unnecessary polyfills and libraries.

asial-matagawa:
@frandiox @misterjunio 
This PR removes unnecessary polyfills and libraries.
Any suggestions?

----
Remove:

- ❌ childNodeRemove.js
  - `Element.prototype.remove` is now supported by all major desktop/mobile browsers including iOS 8-10.

- ❌ classList.js@1.1.20150312
  - `Element.prototype.classList` is now supported by all major desktop/mobile browsers (except for IE11).

- ❌ promise-polyfill@6.0.2+mod
  - `Promise` is now supported by all major desktop/mobile browsers (except for IE11).

- ❌ CustomEvent.js
  - `CustomEvent` is now supported by all major desktop/mobile browsers including iOS 8-10 Safari.

- ❌ MutationObserver@0.7.22
  - `MutationObserver` is now supported by all major desktop/mobile browsers including IE11.

- ❌ winstore-jscompat@399fd9f
  -  Currently Onsen UI is not supporting Windows Desktop/Mobile apps and [this library is no longer needed for Windows 10+](https://github.com/Microsoft/winstore-jscompat/tree/399fd9f0b1d7826e0cbc38e8b4618832f12c4b8e#instructions).

----
Keep:

- FastClick@1.0.6+mod
  - To be removed in #2000.

- microevent.js@47cbc14+mod
  - Non-standard. Must be polyfilled.

- setImmediate@1.0.2+mod
  - Non-standard. Supported only in Edge and IE. Must be polyfilled.

Issue #2001, opened 2017-05-12
Provide ES Modules.


asial-matagawa:

asial-matagawa:
@frandiox (Cc: @anatoo @misterjunio )
We have several options about how to provide a variety of modules.

----
**Option 1:** Provide UMD (bundled), [CommonJS2](https://github.com/webpack/webpack/issues/1114) (bundled) and ESM (bundled).
(Example: [Vue](https://unpkg.com/vue@2.3.3/))

This option is a straightforward solution.
In this option, we have a `package.json` like:

```js
unpkg: 'js/onsenui.js',
main: 'js/onsenui.common.js',
module: 'js/onsenui.esm.js',
```

Non-ES2015 users can use `onsenui.js` with `<script>`.
CommonJS2 users can use `onsenui.js` or `onsenui.common.js` as a CommonJS 2 module.
ES Modules users can use `onsenui.esm.js` as a ES Module.

Note: `unpkg` property prevents that https://unpkg.com/onsenui is resolved to `js/onsenui.common.js`.

----
**Option 2-a:** Provide UMD (bundled) and ESM (bundled) with the extention `.js`.

Maybe [CommonJS2](https://github.com/webpack/webpack/issues/1114) will be replaced by ES Modules in the future, so I think we should not stick to providing a CommonJS2 module.
In this option, we have a `package.json` like:

```js
main: 'js/onsenui.js',
module: 'js/onsenui.esm.js',
```

Non-ES2015 users can use `onsenui.js` with `<script>`.
CommonJS2 users can use `onsenui.js` as a CommonJS2 module.
ES Modules users can use `onsenui.esm.js` as a ES Module.

----
**Option 2-b:** Provide UMD (bundled) and ESM (bundled) with the extension `.mjs`.

According to [this doc by Node.js](https://github.com/nodejs/node-eps/blob/5dae5a537c2d56fbaf23aaf2ae9da15e74474021/002-es6-modules.md#5-semantics) and [this post](http://2ality.com/2017/05/es-module-specifiers.html), it seems that Node.js might use `.mjs` as a default extension for ES Modules in the future.
So it might be better to follow this new specification (but personally I think `.mjs` is not an ideal solution...).

In this option, we have a `package.json` like:

```js
main: 'js/onsenui.js',
module: 'js/onsenui.mjs',
```
or
```js
main: 'js/onsenui',
```

Non-ES2015 users can use `onsenui.js` with `<script>`.
[...]

PR #2008, opened 2017-05-17
feat(onsenui): `onsenui` is now available as ES Modules.

asial-matagawa:
Related issue: #2001

@frandiox @misterjunio 
This PR enables users to use `onsenui` as ES Modules by providing the raw source codes of Onsen UI and `module` property in `package.json`.
I set `core-src/setup.js` as the entry point of ES Module since we already have `css-components-src` as the name for the raw source code directory of CSS Components.

----
**Usage:**

If a user writes `entry.js` as follows,
```js
import ons from 'onsenui';

ons.disableAutoStyling();

```
and run `webpack entry.js bundle.js`, the user will get:
```sh
$ webpack entry.js bundle.js
Hash: 44e9014d7fd88990db94
Version: webpack 2.5.1
Time: 17127ms
    Asset     Size  Chunks                    Chunk Names
bundle.js  1.67 MB       0  [emitted]  [big]  main
  [70] ./~/onsenui/core-src/elements/ons-toolbar.js 16.5 kB {0} [built]
  [76] ./~/onsenui/core-src/setup.js 9.26 kB {0} [built]
  [78] ./entry.js 238 bytes {0} [built]
 [154] ./~/onsenui/core-src/elements/ons-row.js 10.8 kB {0} [built]
 [155] ./~/onsenui/core-src/elements/ons-select.js 17.8 kB {0} [built]
 [156] ./~/onsenui/core-src/elements/ons-speed-dial-item.js 11.5 kB {0} [built]
 [157] ./~/onsenui/core-src/elements/ons-speed-dial.js 21.6 kB {0} [built]
 [158] ./~/onsenui/core-src/elements/ons-splitter-content.js 16.3 kB {0} [built]
 [159] ./~/onsenui/core-src/elements/ons-splitter-mask.js 11.8 kB {0} [built]
 [160] ./~/onsenui/core-src/elements/ons-splitter-side.js 38.7 kB {0} [built]
 [164] ./~/onsenui/core-src/elements/ons-switch.js 21.6 kB {0} [built]
 [166] ./~/onsenui/core-src/elements/ons-template.js 12.6 kB {0} [built]
 [170] ./~/onsenui/core-src/elements/ons-toast/index.js 19.9 kB {0} [built]
 [172] ./~/onsenui/core-src/elements/ons-toolbar-button.js 13.4 kB {0} [built]
 [177] ./~/onsenui/core-src/ons/ons.js 21.2 kB {0} [built]
    + 182 hidden modules
```

----
**Known issue:**
By default, webpack 2 automatically injects `setImmediate` polyfill (and `process` polyfill + timers polyfill) unless the user set `node` option in `webpack.config.js` to suppress the automatic polyfill injection.
This results in duplicate `setImmediate` polyfills in the same bundle (one is `core-src/polyfills/setImmediate@1.0.2+mod/setImmediate.js` and the other one is injected by webpack 2) and makes the bundle a bit bigger.
(But since both of two `setImmediate` polyfills are loaded after `core-src/polyfills/MutationObserver@0.7.22/MutationObserver.js` (which also polyfills `setImmediate`), the problem explained in https://github.com/OnsenUI/OnsenUI/pull/2005#issuecomment-301691981 cannot occur)

At this moment, I have no idea about how to deal with problem...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ dev
  * core: Added new method `ons.createElement(...)` that allows creating new elements from templates or inline HTML.
  * core: Added new method `ons.createElement(...)` that allows creating new elements from templates or inline HTML ([#1941](https://github.com/OnsenUI/OnsenUI/issues/1941)).
  * core: Added `ons-card` element.
+ * core: `onsenui` is now available as ES Modules.
  * core: A fake device back button event is now fired on ESC press.
  * ons-input: Added styling support for `type='search'`.
  * angular1: Added `ons-action-sheet` bindings.
--- a/gulpfile.babel.js
+++ b/gulpfile.babel.js
@@ -469,6 +469,10 @@ gulp.task('prepare', ['html2js'], () =>  {
       .pipe(gulp.dest('build/css/'))
       .pipe(gulpIf(CORDOVA_APP, gulp.dest('cordova-app/www/lib/onsen/css'))),
 
+    // ES Modules (raw ES source codes)
+    gulp.src('core/src/**/*')
+      .pipe(gulp.dest('build/core-src/')),
+
     // angular.js copy
     gulp.src('bindings/angular1/lib/angular/*.*')
       .pipe(gulp.dest('build/js/angular/')),
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
     "test": "gulp test --disable-warnings"
   },
   "main": "js/onsenui.js",
+  "module": "core-src/setup.js",
   "typings": "js/onsenui.d.ts",
   "files": [
     "bower.json",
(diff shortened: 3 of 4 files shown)


frandiox:
@asial-matagawa Thanks a lot for the research! I think including `setImmediate.js` twice is not a big issue since it is a small polyfill (3KB minified). Does this finally enable tree shaking for the components?

asial-matagawa:
@frandiox 
Yes, users can remove unused objects in `onsenui` when building a bundle.
But I think tree shaking for `onsenui` is not as effective as we expect because `onsenui` provides only one object (`default`).

I think tree shaking is effective only when a module provides many objects and a user picks few objects from the module like:

[module.js]
```js
export const a = 1; // removed by tree shaking
export const b = 2; // removed by tree shaking
export const c = 3; // removed by tree shaking
export const d = 4;
export const e = 5; // removed by tree shaking
export const f = 6; // removed by tree shaking
```

[entry.js]
```js
import {d} from './module.js';
```

Sorry if my understanding is wrong.

asial-matagawa:
I will merge this for releasing `2.3.0` for now.
If something happens, I will revert this merge commit.

PR #2010, opened 2017-05-18
refactor(core): Rename `init` to `_init`.

asial-matagawa:
@frandiox @misterjunio 
All the custom element definition classes in the core have `init` method, but it is used as a private method.
Also, @anatoo said that `Maybe we forgot to add the underbar` in Dec 18, 2016.
So I renamed it to `_init`.

I will merge this, but if `init` is a meaningful name I will revert this PR.

### DIFF
--- a/core/src/elements/base/base-dialog.js
+++ b/core/src/elements/base/base-dialog.js
@@ -37,7 +37,7 @@ export default class BaseDialogElement extends BaseElement {
     this.style.display = shouldShow ? 'block' : 'none';
   }
 
-  init() {
+  _init() {
     this._visible = false;
     this._doorLock = new DoorLock();
     this._boundCancel = () => this._cancel();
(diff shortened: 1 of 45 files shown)


frandiox:
@asial-matagawa Actually, I think we were using `init` just due to some fix related to WebReflection's polyfill. Now that we changed to the other polyfill maybe we can use `constructor`. Do you think it is possible?

asial-matagawa:
@frandiox 
>Actually, I think we were using init just due to some fix related to WebReflection's polyfill.

Yes, if we use WebReflection's polyfill, we cannot use `constructor` and have to use a class method, but its name is not restricted to a certain name. So I renamed it to `_init`.

>Now that we changed to the other polyfill maybe we can use constructor. Do you think it is possible?

Yes, it's possible.
I will try to remove `_init`.

PR #2012, opened 2017-05-18
Release `onsenui@2.3.0`.

asial-matagawa:
@OnsenUI/owner


asial-matagawa:
@frandiox 
I think you already know but I found that this line in `ascend-animator.js`causes the forced reflow.
https://github.com/OnsenUI/OnsenUI/blob/master/core/src/elements/ons-toast/ascend-animator.js#L45

https://github.com/OnsenUI/OnsenUI/blob/master/core/src/elements/ons-toast/ascend-animator.js#L32 says `// Calculated during the first animation` but animators are created in https://github.com/OnsenUI/OnsenUI/blob/master/core/src/elements/base/base-dialog.js#L111 each time `BaseDialog#_setVisible` called.
Do you think we can completely suppress this forced reflow?

frandiox:
@asial-matagawa Oh, I think we should only create a new animator if the animation changed. I want to make a small refactor of that animator when I get some free time anyway. I'll try to have a look at it soon.

PR #2018, opened 2017-05-20
remove prop-types warning message

airamrguez:


asial-matagawa:
@airamrguez 
I checked the problem and it was reproduced on my environment.


I will merge this and release `react-onsenui@1.3.1`. Thank you!

PR #2020, opened 2017-05-22
Vue scroll fix

frandiox:
@asial-matagawa What do you think about mapping `pageElement.scrollTop` to `pageElement._contentElement.scrollTop` (81bf4e9)? I think we should provide a cleaner API for our own bindings even if it is not documented. I don't like too much accessing "private" variables (like `_contentElement`) from the bindings. Other solution would be `pageElement.content.scrollTop`.

### DIFF
--- a/bindings/vue/src/components/VOnsNavigator.vue
+++ b/bindings/vue/src/components/VOnsNavigator.vue
@@ -44,21 +44,22 @@
           this.$children[i].$el.style.visibility = visibility;
         }
       },
-      _reattachPage(pageElement, position = null) {
+      _reattachPage(pageElement, position = null, scrollTop = 0) {
         this.$el.insertBefore(pageElement, position);
+        pageElement.scrollTop = scrollTop;
         pageElement._isShown = true;
       },
       _redetachPage(pageElement) {
         pageElement._destroy();
         return Promise.resolve();
       },
-      _animate({ lastLength, currentLength, lastTopPage, currentTopPage}) {
+      _animate({ lastLength, currentLength, lastTopPage, currentTopPage, lastScrollTop }) {
 
         // Push
         if (currentLength > lastLength) {
           let isReattached = false;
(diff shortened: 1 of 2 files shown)


asial-matagawa:
@frandiox 
`pageElement.scrollTop` feels a bit confusing to me since `pageElement` already has its original `scrollTop`.
If it's acceptable, I prefer `pageElement.content.scrollTop`...

frandiox:
@asial-matagawa `pageElement`'s real scrollTop will always be 0, so I don't think it's very interesting. In the past, for example, we were mapping `pageElement`'s style attribute to `.page__background` for the same reason. Do you think it is acceptable or still prefer `pageElement.content`?

asial-matagawa:
@frandiox 
>In the past, for example, we were mapping pageElement's style attribute to .page__background for the same reason.

I see, in that case `pageElement.scrollTop` is better.
Thank you for the PR!

frandiox:
@asial-matagawa I will merge this now. In any case, this is just for internal use so we can change it later.

t2thec:
Hey. Sorry to add to this merge.

I am wanting to add a 'scroll to top' feature (I'm also using the Vue incarnation of Onsen). I can't seem to find any `document.querySelector('#PAGE .SELECTOR').scrollTop` that produces a value greater that `0` that I can then affect.

Any ideas what I am missing? Which element is actually scrolling?

Cheers

frandiox:
@t2thec I think you are looking for `.page__content` selector. In any case, after this was released you can do it directly in the page element (not documented): `pageElement.scrollTop`. Hope it helps!

t2thec:
Thanks @frandiox. Thats is what I had figured.

`document.querySelector('.page__content').scrollTop` just gives me `0` even though I am half way down the page. I'll take another look. I guess I am picking the wrong `.page__content` or something.

Cheers

PR #2023, opened 2017-05-22
fix(ons-splitter): Reveal animation glitch.

frandiox:
@asial-matagawa This should fix the black glitch :smile:


asial-matagawa:
@frandiox 
Thank you!
I will take a look and merge thsi.

asial-matagawa:
@frandiox 
I have tested it and the black glitch disappeared, but a white glitch sometimes appears on Android Chrome 58.
If you have no Android devices now, I think it would be better to concentrate on some other tasks for now.

frandiox:
@asial-matagawa I found that issue, it's something different. This is related to a `visibility = 'hidden'` after the menu is closed. That is necessary because if there are 2 menus with reveal animation, both of them could be visible under some circumstances when swiping one side...
I'll find a different workaround for that :sweat_smile:

asial-matagawa:
@frandiox 
I see, thank you!

frandiox:
@asial-matagawa I've removed `visibility` property and the glitch should be gone now. Instead, I'm changing the `z-index` property of the splitterSides, so the visible one will be displayed over the hidden one. In order to do that I had to adjust all the splitter `z-index` (content, mask, etc).

I cannot see glitches on Firefox or Chrome 58 anymore :)

asial-matagawa:
@frandiox 
I tested the new revision, and it worked fine!

I will merge this, thank you!

PR #2024, opened 2017-05-22
fix(ons-action-sheet): Glitch when hiding on iOS.

frandiox:
@asial-matagawa The bug on iOS 8.0 was visible when moving things inside the view, right? Not when they are moving outside the view. This should fix a [glitch](http://tutorial.onsen.io/?framework=vanilla&category=reference&module=action-sheet) and hopefully won't break the iOS 8.0 thing again.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ dev
 
 ### Bug Fixes
 
+* ons-action-sheet: Fixed glitch when hiding on iOS.
 
 ### Misc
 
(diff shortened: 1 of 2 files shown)


asial-matagawa:
@frandiox 
Yes, I meant that [glitch](http://tutorial.onsen.io/?framework=vanilla&category=reference&module=action-sheet).

I will test it on the real devices and merge this.

asial-matagawa:
@frandiox 
I have checked that the glitch no longer appears on iOS 8.0 🎉 
I will merge this.

frandiox:
@asial-matagawa Thanks!

PR #2042, opened 2017-05-29
chore(unit): Use headless Chrome for unit testing.

frandiox:
:raised_hands: 

@asial-matagawa Can you test this?

### DIFF
--- a/core/src/ons/platform.spec.js
+++ b/core/src/ons/platform.spec.js
@@ -124,9 +124,10 @@ describe('ons.platform', () => {
   });
 
   describe('#isChrome()', () => {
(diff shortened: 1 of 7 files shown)


asial-matagawa:
@frandiox 
I have tested it on my environment (macOS Chrome 58 [latest]) and it failed.


It seems that headless Chrome is shipped in Chrome 57+ in the case of Linux, but in the case of macOS it is shipped in Chrome 59+.
I will implement `local_chrome_headless` as a workaround and after that I will merge this.

frandiox:
Ah! So that was it! I know officially the said v59, but I was surprised it was working on my machine with v58. Thanks for checking!

PR #2043, opened 2017-05-29
Split ons-input

### LINKED ISSUE #2013: Splitting "ons-input" component.
Summary:

We should split `ons-input` component into several components.

Background:

We have `ons-input` component to handle user input  and developers can switch appearance and behavior by setting `type` attribute.

```html
<ons-input type="text" />
<ons-input type="checkbox" />
<ons-input type="search" />
```

But this does not make sense in Onsen UI because some components have unique feature (e.g label feature on `float` attribute and type=text) and these should not be mix together in `ons-input`. Current design of `ons-input` make hard to add unique feature on individual `ons-input` appearance.

Also `ons-input` name make hard to find when developers want to find Onsen UI checkbox element or radio button element or search input element.

So I think we should make individual input components as below like `ons-range` component and deprecate `ons-input` component.

*  `ons-text-input` that accept normal text, password and email format. 
*  `ons-checkbox`
*  `ons-radio-button`
*  `ons-search-input`

@frandiox @misterjunio @asial-matagawa @masahirotanaka Any thought?

frandiox:
This will close #2013 

@anatoo @asial-matagawa @misterjunio @masahirotanaka 

This PR splits the current `ons-input` into 4 different elements:

- `ons-input`: It's now a simple text input. Since it can be `type="password"` and many others, I think `ons-input` is a better name than `ons-text-input`.
- `ons-checkbox`: Old `ons-input[type=checkbox]`. I guess there is no discussion about the name.
- `ons-radio`: Old `ons-input[type=radio]`. I think it's easier to migrate `ons-input type="radio"` to `ons-radio` than `ons-radio-button`, and it is also shorter (things like `v-ons-radio-button` is quite long).
- `ons-search-input`: Old `ons-input[type=search]`. I chose the long version over `ons-search` because the latter is a bit misleading and we might implement an extra `ons-search` in the future.

If you prefer any other naming, please leave your opinion.


frandiox:
@asial-matagawa @misterjunio Can any of you test all of this on Safari and update Angular2 bindings? I will implement soon the other bindings.

misterjunio:
@frandiox everything looks good on the desktop and iOS 10. I think I can update Angular2 👍

misterjunio:
@frandiox actually as far as I investigated the components work as-is in the Angular2 bindings, no need for new directives, maybe @asial-matagawa has a better understanding and can confirm that.

asial-matagawa:
@misterjunio 
Thank you for investigating!
I think we can use basic features of `ons-checkbox` and `ons-radio` without directives, but for advanced ones we might need directives.

I will deal with this problem in #2033.

frandiox:
@misterjunio @asial-matagawa I think React + Vue + Angular1 bindings are also finished :)

asial-matagawa:
@frandiox 
Thank you!

frandiox:
I noticed that we had some DOM elements in `ons-input` related to something deprecated. I've simplified the DOM elements and the code in general in 6da261e . It shouldn't break anything but it's better if we can test it.

frandiox:
@anatoo I've simplified `ons-switch` in c167bf5 , but it has some minor differences. The `checked` attribute is not updated to the inner input, for example. I don't think this was necessary but we can still undo the commit if you think it was important. Native inputs do not react on `checked` attribute changes so I didn't see the point. Now `ons-switch` should behave like `ons-checkbox` and be easier to maintain... hopefully :sweat_smile:

anatoo:
@frandiox Got it. It should be OK.

core/src/elements/ons-input.js
```diff
@@ -126,6 +126,11 @@ export default class InputElement extends BaseInputElement {
       this._input.addEventListener('input', this._boundOnInput);
       this._input.addEventListener('focusin', this._boundOnFocusin);
     });
+
+    const type = this.getAttribute('type');
+    if (['checkbox', 'radio'].indexOf(type) >= 0) {
+      console.warn(`Warn: <ons-input type="${type}"> is deprecated since v2.4.0. Use <ons-${type}> instead.`)
```
asial-matagawa: I guess `util.warn` is a bit better?

frandiox: I forgot we had that utility :sweat_smile:

PR #2048, opened 2017-06-03
Bugfix/relative img paths and angular 2 update

sfstar:
I recently found out that the ons-card visual styling isn't available in the rc.6 version of the angular2 binding.
While implementing a custom upgrade to the newest onsen-ui version I found out that some of the used img paths are incorrect.
Since the select.ccs and search-input.css both use resources from the img while being located in the components directory.

Directory structure:
-css-components
-----src
---------components
-------------search-input.css
-------------select.css
---------img
-------------android-search-input-icon.svg
-------------ios-search-input-icon.svg
-------------select.arrow.svg

so the paths should be relative e.g.
"../img/<resource>"

instead of
"img/<resource>"

furthermore, I upgraded angular to use the newest onsen-ui (core) version.
So, that angular2 binding users can benefit of the ons-card component.

### DIFF
--- a/bindings/angular2/package.json
+++ b/bindings/angular2/package.json
@@ -1,6 +1,6 @@
 {
   "name": "angular2-onsenui",
-  "version": "1.0.0-rc.6",
+  "version": "1.0.0-rc.7",
   "private": false,
   "author": "Mitsunori Kubota <mitsunori@asial.co.jp>",
   "scripts": {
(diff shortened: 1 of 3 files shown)


asial-matagawa:
@sfstar 
Thank you for the PR!
`img/~~~~` is not a relative path but a module path, and actually `../img/~~~~` breaks Onsen UI in some environments.

About `angular2-onsenui`, I am working on a new version of it in #2033.
Please be patient for a while.

@anatoo 
What do you think?
I think `img/~~~~` is weird since when we use the css files directly from browsers without bundlers, the css files should have relative paths.

PR #2051, opened 2017-06-06
fix(action-sheet): Add `!important` to protect `display: none;` from `onsenui.css`.

asial-matagawa:
@anatoo @frandiox
I noticed this problem while preparing Action Sheet example for `ngx-onsenui`.
Is it enough?


asial-matagawa:
@anatoo said that `onsenui.css` should be loaded **before** `onsen-css-components.css`.

Closing this.

PR #2053, opened 2017-06-06
feat(core): Support HTMLTemplateElement.

### LINKED ISSUE #2056: Specs: HTML Template, cache and page life-cycle hooks
Related issues: #1831 #1821 #1814 #1573
PRs: #2053

@anatoo @asial-matagawa @misterjunio @masahirotanaka 

### HTML Templates

All the platforms that we support already include `<template>` element so we should make Onsen UI compatible with it. We can leave `<ons-template>` as is for backward compatibility but prefer `<template>` element from now on. For older platforms, either Crosswalk or a simple [polyfill](https://github.com/webcomponents/template) enables `<template>` element.

```html
<template id="page.html">
  <ons-page>...</ons-page>
</template>
```

### Cache

Right now we are caching only "local" templates (included in `index.html`). We could also cache external files the first time Onsen UI reads them. Also, we can provide an optional `ons.preload([...]);` method that requests those templates and caches them. This would reduce the delay when pushing new pages located outside `index.html`.

### Page life-cycle hooks

Using `<template>` allows `<script>` tags inside it that run when the element is attached. This means that a page can be initialised like this:

```html
<template id="page1.html">
  <ons-page id="p1">
    Content

    <script>
      (function() {
         var page = document.querySelector('#p1');
         // initialise page
      })();
    </script>
  </ons-page>
</template>
```

This would require closures to avoid global pollution and could lead to timing issues with custom elements. Even though there are fixes for these issues, I think it might be better to just provide "life-cycle" hooks:

```html
<template id="page1.html">
  <ons-page id="p1">
    Content

    <script>
      document.querySelector('#p1').init = function() {
        // initialise page
      };
      // same for .show, .hide...
    </script>
  </ons-page>
</template>
```

These hooks will run at the same time as the events are fired.

For the naming, I think `init` could work, but if we want to be consistent with other components perhaps `onInit`, `onShow`, etc. are better (we already have `onAction`, `onInfiniteScroll`, etc.).

I started testing this in #2053.
Opinions?

frandiox:
Closes #2056

@misterjunio ~Can you test this on iOS 8.0 or 8.4 (`examples/test-navigator/index.html`)? I think `importNode` cannot be used with templates in <=8.4, but it would be better to check~ I misunderstood the issue, we don't need to run `importNode` on "templates" but on "template content" (DocumentFragment). It should be fine.

@asial-matagawa ~Are you aware of any difference between `importNode` and `cloneNode`? I added some code to `ons-page` to delay the script run but it only works when using `importNode`. Looks like when using `cloneNode` the script tag in the template runs before page's constructor.~ `cloneNode` does not reset the "already started" flag of `script` tags, so it cannot be used. `importNode` works well.

~In any case, even if we need to use `cloneNode` it can work, we just need to make some extra modifications to `pageLoader`.~

By the way, all the tests pass when using `<template>` for ons-navigator :)


frandiox:
@asial-matagawa @misterjunio Actually, I just thought about something:

```
<template id="page1.html">
  <ons-page id="p1">
    Content

    <script>
       var page = document.getElementById('p1'); // or myNavigator.topPage;
       page.initHook = function() {
           // init page!
       };
       page.showHook = ...
    </script>
  </ons-page>
</template>
```

Then we can run `initHook` at the same time as the `init` event is fired. What do you think about it?

misterjunio:
@frandiox I think the lowest version we have here is `8.4.1`, seems to be working flawlessly there. About the hooks inside the templates I think it's a very good idea. Maybe I would remove `hook` from the naming though, `page.init` seems intuitive for me.

frandiox:
@misterjunio Yes yes, I just added "hook" to make it explicit. I'll open a specs issue and we can decide there. Thanks for testing!

frandiox:
This is finished according to #2056 specs. Changes on this should be easy to make.

misterjunio:
Passed all manual tests 👍

asial-matagawa:
🎉

frandiox:
@misterjunio Thanks for testing!

PR #2057, opened 2017-06-07
Refined appearances of checkbox and radio-button components on iOS.

anatoo:
Refined appearances of iOS checkmark. I will merge this changes if there is no problem.

Old:
image
image

New:
image
image


misterjunio:
Looks good in iOS 8, 9 and 10 👌

anatoo:
@misterjunio Thanks! I merged.

PR #2059, opened 2017-06-08
Perf toast ascend animator.

frandiox:
This needs to be tested in different versions of Android.

@asial-matagawa There shouldn't be more reflows. I realized that the toast is always 1 line... so the heights are always 48 and 64 pixels, no need to compute the height during animation :sweat_smile:
Also, this now always uses `translate3d` for the animations.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ dev
 
  * css-components: Refined the appearance of `button` on Android.
  * core: Updated `font-awesome` icon library.
+ * ons-toast: Improved 'ascend' animator performance.
 
 ### BREAKING CHANGES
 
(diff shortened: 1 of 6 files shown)


asial-matagawa:
@frandiox 
Thank you!
I have no enough time to test this due to some other tasks, but I will check this as soon as possible.

misterjunio:
After trying really hard with the example we have I was able to "break" the toast in both platforms by tapping repeatedly and really fast in the toggling button (at some point it would start hiding the message after opening) but I think it's ok and performance wise really good. No reflows.

frandiox:
@misterjunio I guess it would need some kind of `_isRunning` like the navigator animations, but it's very unlikely that this happens in a real app anyway. Thanks for testing!

PR #2067, opened 2017-06-14
fix(tabbar): Fixed inconsistent appearances of material-tabbar components.

anatoo:
This PR fix material tabbar's white border as the below.

image


asial-matagawa:
👍

PR #2069, opened 2017-06-14
fix(css-components): Fixed incorrect image path.

anatoo:
This fix #2048. I will merge this changes.


asial-matagawa:
Note: Monaca Localkit/CLI templates needs changes for this fix because this change will break the current templates.

PR #2094, opened 2017-06-23
Remove redundant `-` from all modifiers

anatoo:
@asial-matagawa @frandiox @misterjunio I renamed some modifiers have redundant `-` on `ons-button`. I think all modifiers should not have `--` anymore. Should we merge this changes? Let me know what do you think.

Changes: 

  * `large--cta` => `large-cta`
  * `large--quiet` => `large-quiet`
  * `material--flat` => `material-flat`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@ v2.4.2
 
  * ons-fab: Added new appearances for iOS.
 
+### BREAKING CHANGES
+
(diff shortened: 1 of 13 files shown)


masahirotanaka:
I think it a way too big change that affects all users prevent upgrading. Let discuss formatting css class names when we do a major version update.

sbaechler:
Maybe better use two separate css classes since those modifiers are independent. E.g. `.button--large` and `.button--quiet` instead of `.button--large--quiet`.

PR #2078, opened 2017-06-19
Add Lift Animation to ons-modal element

porchdog:

PR #2097, opened 2017-06-27
Modal lift

frandiox:
Continues #2078


frandiox:
@asial-matagawa @anatoo This animation shows the scrollbar. I think the only way to avoid it is to apply `overflow: hidden` to `<body>` or change the DOM structure of `ons-modal`. I guess the first solution is simpler, do you have any objection?

anatoo:
I have no objection.

asial-matagawa:
Sorry for the delay. I am not sure if `overflow: hidden;` causes some issue or not, but if that is the only way, let's go with the solution. 👍

PR #2099, opened 2017-06-29
Page background and iOS slide refactor

frandiox:
@anatoo Can you please have a look at these 2 commits and tell me what you think?
Do you remember what is the `margin-top: -1px` for exactly?

### DIFF
--- a/core/css/common.css
+++ b/core/css/common.css
@@ -96,11 +96,12 @@ ons-bottom-toolbar {
   -ms-touch-action: pan-y;
 }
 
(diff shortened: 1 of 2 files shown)


anatoo:
@frandiox Good work!

>Do you remember what is the margin-top: -1px for exactly?

In days of old, I added this line for appearances of ons-toolbar on running iOS "slide" animation. But I forgot about the details. It's OK to remove if you confirm the animation is fine.

frandiox:
@anatoo Thanks! I couldn't see any issue in the examples we have so I'll merge this.

PR #2104, opened 2017-07-03
Swipe to pop feat

frandiox:
@asial-matagawa @anatoo @misterjunio @masahirotanaka 

This PR implements "swipe to pop" feature for iOS navigation. It has the same API as `ons-splitter-side`, i.e. `swipeable`, `swipe-threshold` and `swipe-target-width` attributes. Actually, all the implementation can be shared with `ons-splitter-side` and I'd like to refactor that component a bit to avoid code duplication. Forcing this feature on Android is possible by specifying `swipeable="force"`.

You can test it in the normal navigator example, just include `swipeable` attribute.
This still needs docs and support for React and Vue bindings and tests with carousel and other components.

Any issue/opinion/suggestion?


frandiox:
More info about swiping bugs in #2107.

asial-matagawa:
@frandiox 
I tested 3a8285f70836143b5e8c1603dee5abe183028552 on macOS Safari 10 and it worked fine!
I have no opinion/suggestion about this.

PR #2105, opened 2017-07-04
chore(vue): Minify and source maps.

frandiox:
@asial-matagawa This generates `vue-onsenui.js`, `vue-onsenui.js.map`, `vue-onsenui.min.js` and `vue-onsenui.min.js.map`. Do you think it is fine like this or should remove `vue-onsenui.min.js.map`?

### DIFF
--- a/bindings/vue/webpack.config.prod.js
+++ b/bindings/vue/webpack.config.prod.js
@@ -5,13 +5,16 @@ var webpack = require('webpack')
 const pkg = JSON.parse(fs.readFileSync('package.json', 'utf-8' ));
 
 module.exports = {
-  entry: './src',
+  entry: {
+    'vue-onsenui': './src',
+    'vue-onsenui.min': './src'
+  },
   output: {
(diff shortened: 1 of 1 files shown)


asial-matagawa:
@frandiox 
Thank you!

>Do you think it is fine like this

Yes, it is fine!
For users who use `vue-onsenui` as an npm package, it does not matter whether the sourcemap is inline or not.

>should remove vue-onsenui.min.js.map?

I think we should keep `vue-onsenui.min.js.map` since we can provide the sourcemap of `vue-onsenui.min.js` for CDN users if they write `<script src="https://unpkg.com/vue-onsenui/dist/vue-onsenui.min.js"` (Unfortunately, we cannot it for those who directly download `vue-onsenui.min.js` since they probably ignore `vue-onsenui.min.js.map`).

frandiox:
@asial-matagawa Thanks, I'll do the same for React then.

Issue #604, opened 2015-04-17
PullHook animation freezes for a moment when dragging back


frandiox:
Drag PullHook down  and then up slowly again without releasing it:
http://codepen.io/onsen/pen/WbJogM

Tested in Chrome and Firefox.

Edit: Doing it several times in a row without releasing PullHook increase the error.

argelius:
Thanks @frankdiox !

misterjunio:
Fixed in v2 😄

PR #2107, opened 2017-07-06
Cross swipe

frandiox:
@anatoo @asial-matagawa @misterjunio This should fix most of the issues for swipeable components. Please have a look and play with it a little bit (mostly on Safari/devices). There is an example in `examples/swipe/index.html` that combines all of them.

Basically, the first element takes the event if it can consume it. Otherwise, it bubbles. There is, however, the HammerJS' `dragLockToAxis` with an error margin of 25px and this should also work when there are no parent components that can consume the event.

Apart from that, I removed events like `dragright dragleft dragup dragdown` and `swipe swiperight swipeleft swipeup swipedown` in favor of `drag dragstart dragend`. The carousel was using all of them and, as a result, it was calling the handlers more times than needed. @anatoo Do you think this can be an issue? I think `drag` should be enough, right?

### DIFF
--- a/bindings/angular2/examples/pull-hook.ts
+++ b/bindings/angular2/examples/pull-hook.ts
@@ -2,6 +2,8 @@ import {
   Component,
   OnsenModule,
   NgModule,
+  ViewChild,
+  AfterViewInit,
   CUSTOM_ELEMENTS_SCHEMA
 } from '../src/ngx-onsenui';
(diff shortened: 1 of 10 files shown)


frandiox:
Bugs like the one reported in #604 should now be fixed for all the components.

frandiox:
I'm merging this into the other PR to continue with the bindings.

PR #2109, opened 2017-07-12
Added ripple effects on some components.

anatoo:
Summary:

 * Added ripple effect on `ons-toolbar-button` components for material design.
 * Added ripple effect on `ons-alert-dialog` components' buttons for material design.
   * Added `ons-alert-dialog-button` components and added ripple effect.
 * Added ripple effect on `ons-range`, `ons-radio-buton` and `ons-checkbox` components for material design.
 * Fixed some inconsistent appearance on `ons-range` for material design.
 * Added `size` attribute on `ons-ripple` components.

I will merge this changes if there is no problem.


frandiox:
@anatoo Looks great, thanks! Just one thing, the ripple stays visible until you click outside it. Perhaps we should use `:active` instead of `:focus`?

frandiox:
@anatoo I see. It feels weird for me to have it activated all the time after pressing it :confused: 
I think many users will ask for a way to deactivate ripple effect on checkboxes and radios...

anatoo:
I think It's weird a little. But I would like to add ripple effects on focus for now.
If many users complain this behavior after this feature is released, I will make a plan.

PR #2112, opened 2017-07-14
Added important missing methods to the object

raphox:
In the documentation (https://onsen.io/v2/api/vue/$ons.html) we can see other important methods, but in definition of object these methods do not exist.

### DIFF
--- a/bindings/vue/src/index.js
+++ b/bindings/vue/src/index.js
@@ -47,6 +47,7 @@ const install = (Vue, params = {}) => {
    */
   Vue.prototype.$ons = Object.keys(ons)
     .filter(k => [
+      /^is/,
       /^disable/,
       /^enable/,
       /^open/,


frandiox:
@raphox Thanks! The `isWebView()` method is also exposed in `$ons.platform.isWebView()`, but I guess this change is fine. The docs for `$ons` are currently showing the original `ons` object, we need to filter those methods :sweat_smile:

raphox:
Maybe `forcePlatformStyling` is required too.

In my case both is required:

```javascript
// ...
import ons from 'onsenui'
// ...
new Vue({
  el: '#app',
  router,
  store,
  i18n,
  template: '<App/>',
  components: { App },
  beforeCreate () {
    if (!this.$ons.platform.isIOS()) {
      ons.forcePlatformStyling('android')
    }

    /* global StatusBar navigator */
    /* eslint no-undef: "error" */
    this.$ons.ready(function () {
      if (ons.isWebView()) {
        StatusBar.backgroundColorByHexString('#FFAABB')
        navigator.splashscreen.hide()
      }
    })
  }
})
```

frandiox:
@raphox `forcePlatformStyling` is just for changing the platform in demos. You can use `$ons.platform.select('android')` instead.

PR #2115, opened 2017-07-19
Reveal, PullHook and Swipe to pop fixes for 2.5.0

frandiox:
@asial-matagawa Due to some changes in animations, the side menus must be hidden for reveal animator, otherwise, they will be partially visible when pushing pages in the navigator. I remember there was a white glitch when using `visibility: hidden` but I couldn't find it on Chrome or Firefox with this `display: none`. Do you think it is fine?

### DIFF
--- a/core/css/common.css
+++ b/core/css/common.css
@@ -105,6 +105,12 @@ ons-bottom-toolbar {
   overflow: visible;
 }
 
+.page.overflow-visible .page__content.content-swiping,
+.page.overflow-visible .page__content.content-swiping .page,
+.page.overflow-visible .page__content.content-swiping .page__content {
+  overflow: auto;
+}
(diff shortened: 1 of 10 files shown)


asial-matagawa:
@frandiox 
Sorry, I am not sure about it now...
I think this kind of changes always need manual testing in real devices.

@misterjunio 
Actually I was told to stop participating in development now.
Could I ask you to make sure that this change does not cause glitches in Android/iOS devices?

misterjunio:
@asial-matagawa sure, will do

misterjunio:
@frandiox @asial-matagawa tested in all devices, no issues found 👍

asial-matagawa:
Thank you! 👍

PR #2116, opened 2017-07-19
Replaced font icons with SVG icons from "ons-back-button" components.

anatoo:
@frandiox I replaced font icons with SVG icons from "ons-back-button" components and added `svg-inline-loader` on webpack configs. As a result, I removed font-awesome and ionicons dependencies on core.

Could you review this changes?


frandiox:
@anatoo Good job! The only difference I can notice is that in MD, it doesn't have any effect when the back-button is pressed anymore. I think it's not very important, though :+1:

PR #2117, opened 2017-07-20
Added ripple effects for ons-back-button.

anatoo:
Added ripple effects properly for ons-back-button. I will merge this changes if there is no problem.


frandiox:
@anatoo Looks good, thanks!

Issue #2121, opened 2017-07-21
Specs: <ons-segment>


misterjunio:
@OnsenUI/members Here is my proposal for the `<ons-segment>` component.

### DOM structure
Current implementation expects the usage of buttons inside the segment like this:
```html
<ons-segment>
  <button class="button">Label 1</button>
  <button class="button">Label 2</button>
  <button class="button">Label 3</button>
</ons-segment>
```
which is compiled into:
```html
<ons-segment class="segment">
  <div class="segment__item">
    <input type="radio" class="segment__input" name="ons-segment-gen-1" checked id="ons-segment-gen-1-0">
    <button class="button segment__button">Label 1</button>
  </div>
  <div class="segment__item">
    <input type="radio" class="segment__input" name="ons-segment-gen-1" checked id="ons-segment-gen-1-1">
    <button class="button segment__button">Label 2</button>
  </div>
  <div class="segment__item">
    <input type="radio" class="segment__input" name="ons-segment-gen-1" checked id="ons-segment-gen-1-2">
    <button class="button segment__button">Label 3</button>
  </div>
</div>
```
Another option would be to have another component `<ons-segment-item>` inside `<ons-segment>` but it's such a simple component that I feel that's not necessary and this syntax seems to me intuitive and easy to use enough.

### Connection to `<ons-tabbar>`
Currently users can pass a `tabbar` attribute to `<ons-segment>` with the ID of the `<ons-tabbar>` they want to connect the segment. This automatically hides the tabbar and syncs the active button of the segment with the active tab of the tabbar.

Eventually the `<ons-tabbar>` could be found automatically in the DOM but since there may be more than one I think's it's better to just let users pass the ID.

Obviously this is expected to be used in a toolbar otherwise the tabbar pages would hide the segment. By default the styles already work in the `<ons-toolbar>`.

### Attributes
* `tabbar` - ID of the `<ons-tabbar>` to "connect" to the segment
* `active-button` - index of the first active button, only works if there is no connected tabbar (in which case the active tab sets the active button)

### Properties
* `tabbar` - to provide easier access to the connected tabbar

### Methods
* `setActiveButton((index, [options])` - makes button with the specified index active; if there is a connected tabbar it shows the corresponding tab page; in this case animations and other options can be specified by the second parameter
* `getActiveButtonIndex()` - returns button index of current active button; if active button is not found, returns -1

### Events
* `change` - thrown when the active button changes

[...]

PR #2124, opened 2017-07-24
Segment

misterjunio:
@OnsenUI/members here is the PR for working on the `<ons-segment>` component with the specs in #2121.


misterjunio:
@anatoo can I ask for your help with something, please? If you checkout this branch you can see that in all the frameworks I have created an example with `ons-segment` inside `ons-toolbar`. Everything is fine except in Angular 2, where instead of being vertically centered, the segment appears on the top of the toolbar. In the inspector I see that it's getting a different height but the styles are exactly the same in both cases so I can't figure out why that's happening. Do you mind taking a look when you have some time?

anatoo:
@misterjunio Ok, I will take a look today.

anatoo:
@misterjunio I had take a look while ago, but I also can't get why. So I added dirty fix.

PR #2126, opened 2017-07-25
Add swipePop to Navigator

airamrguez:
Add swipeable and swipePop method to Navigator component.

### DIFF
--- a/bindings/react/src/components/Navigator.jsx
+++ b/bindings/react/src/components/Navigator.jsx
@@ -270,6 +270,7 @@ class Navigator extends BasicComponent {
     node.addEventListener('prepop', this._prePop);
     node.addEventListener('postpop', this._postPop);
 
+    node.swipeMax = this.props.swipePop;
     node.onDeviceBackButton = this.props.onDeviceBackButton || this._onDeviceBackButton.bind(this);
 
     if (this.props.initialRoute && this.props.initialRouteStack) {
@@ -404,7 +405,30 @@ Navigator.propTypes = {
    *  [ja][/ja]
    */
   animationOptions: PropTypes.object,
+ 
+  /**
+   * @name swipeable
+   * @type bool|string
+   * @required false
+   * @description
+   *  [en]
+   *  Enables swipe-to-pop functionality for iOS.
+   *  [/en]
+   *  [ja][/ja]
+   */
+  swipeable: PropTypes.oneOfType([PropTypes.bool, PropTypes.string]),
 
+  /**
+   * @name swipePop
+   * @type function
+   * @required false
+   * @description
+   *  [en]
+   *  Function called on swipe-to-pop. Must perform a popPage with the given options object.
+   *  [/en]
+   *  [ja][/ja]
+   */
+  swipePop: PropTypes.func,
   /**
    * @name onDeviceBackButton
    * @type function


frandiox:
@airamrguez Hi! Thanks for the PR. I was checking the Navigator in React and thought that there's no need to change anything (apart from `swipeable` propType, which I forgot) since it has a `postpop` event handler with some logic. Is the page stack inconsistent or anything after swiping?

airamrguez:
@frandiox Hi! I have a wrapper around the Navigator component. So instead of calling methods like pop and push directly in the Navigator, I call them from my wrapper component. On my wrapper component I only need to add:

```

**Comment by frandiox** [comment]
@airamrguez I've checked this again before releasing v1.5.1 but I saw again that `swipePop` is not strictly necessary for a basic usage. What kind of issue does your wrapper fix and how would `swipePop`  prop help you better than `onPrePop` event (**pre**pop)?
I'll wait for this information before merging and releasing the current changes.

**Comment by airamrguez** [comment]
@frandiox I'm going to be very busy this week. I'll try to take a look at it asap. Sorry for the inconvenience

**Comment by airamrguez** [comment]
@frandiox I have a wrapper around the `Navigator` component to achieve 3 things:

1. **Avoid getting the UI unsuful** on some Android devices because **hitting the back button many times very fast** give us the error

> Navigator is already running animation.

2. Keep track of the **active route**. Our app has a side menu and we want to highlight the current section.
3. Keep track of the **navigation depth**. I use this in many ways. For example to generate keys for React components, because the same view can be pushed into the navigation stack many times from different parts of our app.

For example, the popPage method is overloaded like this:
```js
  popPage(...args) {
    if (RoutingNavigator.depth <= 0) {
      return Promise.resolve();
    }
    if (this.nav.isRunning()) {
      // Router in the middle of a transition. Defferring it ...
      const defer = loadingDefers.get(RoutingNavigator.depth);
      if (defer !== undefined) {
        return defer.then(() => {
          this.popPage(...args);
        });
      }
      return Promise.reject('Navigator is already running animation.');
    }
    RoutingNavigator.depth--;
    const deferKey = RoutingNavigator.depth;
    const lastHistoryIndex = Math.max(0, this.history.length - 2);
    const previousRoute = this.history[lastHistoryIndex];
    this.eventEmitter.emit('routechange', previousRoute);

    const unloadDefer = this.nav.popPage(...args).then(() => {
      this.history.pop();
      loadingDefers.delete(deferKey);
    });

    return unloadDefer;
  }
```

Why do I need a `swipePop` method? Because `swipePop` should do at least tasks 2 and 3. It can be implemented like this:

```js
  swipePop = (...args) => {
    this.popPage(...args);
  };
```

frandiox:
@airamrguez Thanks a lot for taking time to write this. I might be wrong but I think all of that can be achieved also with the core events (props in React). For example, `onPrePop` prop gets an `event` that contains `routes: { poppingRoute: {}, routes: [...] }`. Same for `onPostPop` but it contains `poppedRoute` instead of `poppingRoute`. And I guess `event.routes.routes.length` can be used for the navigation depth?

As for the `Navigator is already running animation.` error, I think it shouldn't happen anymore in recent versions since there is a transparent mask during animations which takes extra clicks (so it shouldn't be possible to click on back buttons). It might perhaps happen with the device back button, but it should be possible to catch errors with `this.popPage().catch(...)` anyway since it returns a promise.

I guess we can still add `swipePop` if you think it's necessary for this use case. It should be completely optional, though.

PR #2134, opened 2017-08-01
feat(ons): Add ons.getScriptPage method.

frandiox:
@misterjunio @asial-matagawa It should be a function because we don't document properties in `ons` object. Perhaps `ons.getScriptPage()` would be a better name, what do you think?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ dev
 
  * ons-navigator: Ignore swipes on back buttons.
 
+### Misc
+
+ * ons.getScriptPage: Added new method as a shortcut to get the current page and attach lifecycle hooks.
+
 v2.5.1
 ----
 
--- a/core/src/ons/ons.js
+++ b/core/src/ons/ons.js
@@ -229,7 +229,7 @@ ons.enableAutoStyling = ons._autoStyle.enable;
  * @method forcePlatformStyling
  * @signature forcePlatformStyling(platform)
  * @description
- *   [en]Refresh styling for the given platform.[/en]
+ *   [en]Refresh styling for the given platform. Only useful for demos. Use `ons.platform.select(...)` for development and production.[/en]
  *   [ja][/ja]
  * @param {string} platform New platform to style the elements.
  */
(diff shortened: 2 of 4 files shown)


misterjunio:
@frandiox to be honest the name does not sound too good but truth is I can't come up with a better one 😅  `ons.getScriptPage()` seems ok

frandiox:
@misterjunio Doesn't sound very good but I think it's easy to understand and nobody will use it in a different context...
I always forget including new methods in the d.ts :sweat:

asial-matagawa:
@frandiox @misterjunio 
Thank you for the suggestion.

I think the current implementation could cause a usability issue.
How about doing this?
- If `document.currentScript` is not defined, show `[onsenui] ons.getScriptPage() does not support this browser. Use document.getElementById instead.`
- If `document.currentScript` is defined but it is `null`, show `[onsenui] ons.getScriptPage() must be called in <script> in <ons-page>.`

frandiox:
@asial-matagawa Thanks for the input. `document.currentScript` is actually supported in all the browsers officially supported by Onsen UI. I just added the fallback for those who risk and try to use Android 4 without crosswalk, so I think we shouldn't worry too much about it.

This is why I prefer to call it `ons.getScriptPage()`, because it is obvious that it should only be used for the script tags. Therefore, returning `null` seems enough when there is no page. What do you think?

asial-matagawa:
@frandiox 
>I just added the fallback for those who risk and try to use Android 4 without crosswalk

Thank you!

>so I think we shouldn't worry too much about it.

I think many users still try to use older Android versions (for example, 4.3).
What I concerned about is that `document.currentScript` returns `null` but `document.scripts[document.scripts.length - 1]` does not return `null` even if `ons.getScriptPage()` is called outside `<script>` in older Android browsers.

>Therefore, returning null seems enough when there is no page. What do you think?

It is also fine 👍 
I was just afraid that  `Uncaught TypeError: Cannot read property 'someProperty' of null` could confuse users.

frandiox:
@asial-matagawa 
>document.scripts[document.scripts.length - 1] does not return null

Yes, but `/ons-page/i.test(getCS().parentElement.tagName)` will likely fail so the final result will still be `null`. I'll try to document this enough in the guide as well so nobody uses it outside `script` tags anyway.

asial-matagawa:
@frandiox 
Oh, I see. Thank you for explaining.

PR #2135, opened 2017-08-02
React back button handler improvement

misterjunio:
Extends #1887 to all the components that implement device back button handlers.


frandiox:
@misterjunio @airamrguez Thanks!

PR #2148, opened 2017-08-17
Refined css-component previewer.

anatoo:
@frandiox @asial-matagawa @misterjunio I refined css-components previewer. Would you like to provide any feedback?

Summary:
 * Refined css-components previewer.
   * Reimplemented as a SPA with Vuejs and pagejs.
 * Make iOS components' colors to native-like.

image
image

On mobile:
image


frandiox:
@anatoo Thanks, looks great!

Are you going to implement a tool for changing colors from the previewer and then downloading it? Or that is not planned?

anatoo:
@frandiox No and not planned. This previewer already has CSS live inject on changing css files. So I think that isn't needed.

PR #2156, opened 2017-08-25
fix(css-components): Fixed a collision of "nodivider" modifier and "chevron" modifier.

anatoo:
@frandiox I don't know why there is ".list-item--nodivider:before { border: none; background:none; } ", but I fixed.


frandiox:
@anatoo Thanks!

PR #2175, opened 2017-09-08
fix(react/navigator): remove unnecessary renderPage

emj365:
remove the renderPage executing in navigator which would render page component twice

### DIFF
--- a/bindings/react/src/components/Navigator.jsx
+++ b/bindings/react/src/components/Navigator.jsx
@@ -307,11 +307,10 @@ class Navigator extends BasicComponent {
   render() {
     const {...others} = this.props;
     Util.convert(others, 'animationOptions', {fun: Util.animationOptionsConverter, newName: 'animation-options'});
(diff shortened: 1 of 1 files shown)


frandiox:
@emj365 Thanks! They are already rendered in `componentDidMount` and pushPage-like methods, it's true :)

airamrguez:
@frandiox the line deleted in fact was really useful. We've come back to the initial approximation which I think is wrong. The last time it was changes was in [this pull request](https://github.com/OnsenUI/react-onsenui/pull/139/files). 

So the problem is basically that you have a parent component over the navigator component and a child component inside the navigator (rendered from the `renderPage` prop). If the child updates a parent's prop, then the inner components are not re-rendered and so they are not being updated.

Now my code is broken😔

frandiox:
@airamrguez Oh, I didn't know about that PR. I guess we should revert this change :/

PR #2180, opened 2017-09-13
Refined Android color scheme.

anatoo:
@asial-matagawa @frandiox Refined default Android color scheme. I will merge this changes if there is no problem.

image


asial-matagawa:
@anatoo 
Thank you!
Because this is a kind of breaking change, I will add some documentation about coloring to onsen.io .

PR #2183, opened 2017-09-13
Migrate to CircleCI 2.0

serima:
## Summary

CircleCI runs speed up.
But the CI failed (https://circleci.com/gh/serima/OnsenUI/29) now.

Although I guess, We are using SauceLabs' service to test in Safari, but the access key and token for that are required to be set as environment variables.
Therefore, I think that if the owner is `serima` it will fail the test.

Here is the result of trying and testing it by changing it in `gulpfile.babel.js` as follows.

https://circleci.com/gh/serima/OnsenUI/28 is passed.

```diff
diff --git a/gulpfile.babel.js b/gulpfile.babel.js
index 474844059..e9fdfded6 100755
--- a/gulpfile.babel.js
+++ b/gulpfile.babel.js
@@ -384,7 +384,7 @@

     (async () => {
       const specs = argv.specs || 'core/src/**/*.spec.js'; // you cannot use commas for --specs
-      const browsers = argv.browsers ? argv.browsers.split(',').map(s => s.trim()) : ['local_chrome_headless', 'remote_macos_elcapitan_safari_10'];
+      const browsers = argv.browsers ? argv.browsers.split(',').map(s => s.trim()) : ['local_chrome_headless'];

       let listOfSpecFiles;
       if (argv.separately) { // resolve glob pattern
```

How can I proceed from here?

## Comparison

Due to the above problem, I can not compare strict performance comparison, but I think if the dependency library is cached, the build time will be about half.

Now: over 10 minutes.
Future: about 6 minutes.


serima:
Hi @asial-matagawa, Could you check this? It is probably an environmental variable problem, how can I avoid it?

.circleci/config.yml
```diff
@@ -0,0 +1,132 @@
...
+            cd bindings/vue
+            yarn install --pure-lockfile
+      - save_cache:
+          key: v1-vue-cache-{{ .Branch }}-{{ checksum "bindings/vue/yarn.lock" }}
+          paths:
+            - ./bindings/vue/node_modules
+
+      - run:
+          name: Run OnsenUI Tests
+          command: npm run test:skip-build
```
asial-matagawa: Hi @serima, thank you for your great PR!

I think replacing this line with `gulp test --disable-warnings --skip-build --browsers local_chrome_headless` works well in this case.
`--browsers local_chrome_headless` option suppresses the unit testing on Safari in SauceLabs.

serima: Thank you. I will try it!

serima:
@asial-matagawa It certainly passed CI. Can you merge this in this state once?
After that, it seems good to create a pull request to revert it as OnsenUI.

asial-matagawa:
@serima 
I see, I will merge this PR and after that revert 963d05bb020477c86f205ea8e4970ba41237a91c directly.
Thank you for the PR!

PR #2185, opened 2017-09-14
Migrate to CircleCI 2.0

serima:
In #2183, CI runs as `serima`. it failed.
I changed the setting of CircleCI so I tried to create pull-request again 🙇


serima:
hmm, not running.

PR #2187, opened 2017-09-22
fix (BasicComponent): null check result of findDOMNode

rynomad:
Apologies for the commit message not following proper formatting. 

If updateClasses somehow gets called after a component has unmounted and findDOMNode returns null or undefined, an error invariably gets thrown on the first property lookup of node. This pull request simply adds a small check to return early from the updateClasses function in such a case.

### DIFF
--- a/bindings/react/src/components/BasicComponent.jsx
+++ b/bindings/react/src/components/BasicComponent.jsx
@@ -11,7 +11,9 @@ class BasicComponent extends React.Component {
 
   updateClasses() {
     var node = ReactDOM.findDOMNode(this);
-
+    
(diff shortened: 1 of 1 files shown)


frandiox:
@rynomad Thanks! Did this happen in your app?

rynomad:
yeah. My use case is a bit 'out there'... I'm using OnsenUI with mobx and inferno (with inferno-compat to bridge the react API) and I'm mixing in some components from Blueprint.js. so I'm sure something else was the root cause of the race condition, but I'll find that out the next time it bites me.

PR #2198, opened 2017-09-29
Added dark-theme and enhanced theme previewer.

anatoo:
@asial-matagawa @frandiox 

I added new dark theme. I will merge this changes if there is no problem.

Summary:
 * Added new dark-theme.
 * Added patterns page in the theme previewer.
 * Added theme select control in the theme previewer.

image

image


frandiox:
@anatoo Good job! Can we merge this right before releasing 2.7.0? Otherwise, we might need to revert the changes again for releasing minor versions.

anatoo:
@frandiox Yes, it's OK.

PR #2200, opened 2017-10-03
Fix usage of alpha function in css-components

jalkoby:
The alpha function was not correctly used as the result was produced an invalid css which fails in the scss/sass environment #2199


asial-matagawa:
Hi, @jalkoby.
Thank you for the PR!

@anatoo 
Could you review this change?

anatoo:
@jalkoby Thanks! Merged.

jalkoby:
it was fast 👍

PR #2218, opened 2017-10-13
iPhone X (UI|WK)WebView and Safari support

asial-matagawa:
@anatoo @frandiox @masahirotanaka @andipavllo 

This PR includes a patch for iPhone X.

## Fundamentals

iPhone X has (1) **notch**, (2) **round corners** and (3) **home bar**.
For more details, please refer to official or third-party resources.

If `viewport-fit=cover` setting is not enabled, Onsen UI works completely, but most users (especially Cordova users) will want to enable `viewport-fit=cover`.
If `viewport-fit=cover` setting is enabled, many things will be broken, so in that case we have to deal with the three obstacles in both of (UI|WK)WebView and Safari.

## Requirements for supporting (UI|WK)WebView and Safari

This section assumes environments where `viewport-fit=cover` is enabled.

### (UI|WK)WebView

In **portrait** mode, all of the notch, the round corners and the home bar can collide with the Onsen UI Components, especially top/bottom toolbars, toolbar-less pages, top/bottom tab bars, dialogs and FABs.

In **landscape** mode, all of the notch, the round corners and the home bar can collide with the Onsen UI Components, especially toolbar items, page contents and lists.

### Safari

In **portrait** mode, the notch and the upper round corners are padded with the address bar, and if `ons-page` is used, the bottom round corners and the home bar are also padded with the native tab bar.

In **landscape** mode, the situation is the same as the portrait mode of (UI|WK)WebView except that the address bar can appear.


asial-matagawa:
@anatoo @frandiox 
I created a proposal for this patch.
You need not to read all the sentences but if you notice something critical, I would appreciate if you inform me. Thank you!

## Specification (Proposal)

### Put this patch into `onsen-css-components.css` instead of `onsenui.css`

I think style rules which does not contain certain custom elements name such as `ons-tabbar` should be implemented in CSS layer instead of JS layer.

### Change style of outermost elements by using CSS only

Supporting the notch, the round corners and the home bar is similar to supporting iOS 7+ status bar.

The current auto status bar implementation attaches `status-bar-fill` attribute with JavaScript to `ons-page` and `ons-tabbar` for adding 20px padding to outermost pages/toolbars/tabbars, but I noticed that we can change style of outermost pages/toolbars/tabbars without JavaScript.

For example, if we want to attach `background-color: red;` to outermost `.foobar`, we can use:
```css
.foobar:not(.foobar .foobar) {
  background-color: red;
}
```
But actually using complex selectors in `:not(X)` is not allowed, so we have to write 
```less
.foobar {
  background-color: red;
}
.foobar .foobar {
  // Restore original style which `.foobar` had before
  background-color: transparent;
}
```

So I would like to this way for this patch.

### Use `body[onsflag-iphonex-portrait]` and `body[onsflag-iphonex-landscape]` to enable iPhone X support

If we write style rules as follows and use `body[hoge]`, we can toggle iPhone X support without JavaScript.

```less
body[hoge] .foobar {
  background-color: red;
}
body[hoge] .foobar .foobar {
  // Restore original style which `.foobar` had before
  background-color: transparent;
}
```

### Fill the safe areas using raw values instead of using `constants(safe-area-inset-*)`

Unfortunately **the current iPhone X UIWebView has a critical bug on `constants(safe-area-inset-*)`**.

> 34518947: UIWebView safe area inset constants are not updated on rotation
> https://github.com/lionheart/openradar-mirror/issues/18415

Moreover, **Chrome cannot recognize values like `calc(constant(safe-area-inset-*) + 44px)`.**

Therefore, we should not use `constant(safe-area-inset-*)` until safe area is safely implemented in many devices, and we cannot support devices which have safe areas at this moment.

### Suppress auto status bar fill for iPhone X

Status bar is replaced with top safe area in iPhone X.
We have to suppress auto status bar fill running on iPhone X.

### Not automatically turning on this patch

iPhone X support is not always required by users.
For example, `viewport-fit=cover` is not enabled, Onsen UI must not try to padding safe areas.
Therefore iPhone X support can be manually configured and `ons.platform.isIPhoneX()` should be provided.

```js
// For (UI|WK)WebView
if (ons.platform.isWebView() && ons.platform.isIPhoneX()) {
  document.body.setAttribute('onsflag-iphonex-portrait'); // Required for WebView
  document.body.setAttribute('onsflag-iphonex-landscape'); // Required for WebView and Safari 
}
```

### Not automatically setting `viewport-fit=cover`

This is because basically a library should not pollute global configurations.

asial-matagawa:
## Progress

- [ ] action-sheet.css
- [ ] alert-dialog.css
- [ ] button-bar.css
- [ ] button.css
- [ ] card.css
- [ ] checkbox.css
- [ ] dialog.css
- [ ] fab.css
- [ ] global.css
- [ ] list.css
- [ ] modal.css
- [ ] notification.css
- [ ] page.css
- [ ] popover.css
- [ ] progress-bar.css
- [ ] progress-circular.css
- [ ] radio-button.css
- [ ] range.css
- [ ] search-input.css
- [ ] segment.css
- [ ] select.css
- [ ] switch.css
- [ ] tabbar.css
- [ ] text-input.css
- [ ] textarea.css
- [ ] toast.css
- [ ] toolbar-button.css
- [ ] toolbar.css

asial-matagawa:
## Expected behavior of Action Sheets

### Portrait mode

#### iPhone 8 (iOS 11)

The left/right margins should be 10 pt.
The bottom margin should be 10 pt.

simulator screen shot - iphone 8 - 2017-10-17 at 22 02 32

#### iPhone X (iOS 11)

The left/right margins should be 10 pt.
The bottom margin should be 48 pt, which consists of 34 pt (bottom safe area) + 14 pt (extra bottom margin).

simulator screen shot - iphone x - 2017-10-17 at 22 02 30

### Landscape mode

#### iPhone 8 (iOS 11)

The width should be the same as the portrait mode.
The bottom margin should be 10 pt.

simulator screen shot - iphone 8 - 2017-10-17 at 22 02 06

#### iPhone X (iOS 11)

The width should be the same as the portrait mode.
The bottom margin should be 33 pt, which consists of 21 pt (bottom safe area) + 12 pt (extra bottom margin).

simulator screen shot - iphone x - 2017-10-17 at 22 02 02

anatoo:
>Suppress auto status bar fill for iPhone X
>
>Status bar is replaced with top safe area in iPhone X.
>We have to suppress auto status bar fill running on iPhone X.

Do you say about mock status bar? I think status bar fill is needed for top padding on iPhone X.

asial-matagawa:
@anatoo 
Yes, I mean that.

I added `padding-top: 44px;` into `ons-toolbar` in this patch, so if the current mock status bar (20px) is added it will have 64px. This is not an ideal situation.
Users have to call `ons.disableAutoStatusBarFill()` in iPhone X devices to prevent this situation, but I think we should do that on behalf of users. `Suppress auto status bar fill for iPhone X` means that.

Issue #2237, opened 2017-10-31
React | Page issue: leaking modifiers to grandchild elements


hugofelp:
__Environment__

  
```
[Core]
  onsenui latest
  
[Framework]
  react 15.4.2

[Framework binding]
 react-onsenui latest

[Platform]
  Desktop - OS X 10.11.6 64-bit

[Browser]
  Desktop - Chrome 61.0.3163.100 on OS X 10.11.6 64-bit
```

__Encountered problem__
  Nested Page components accumulate modifiers from parent Page. The code provided in the example will render the following for the inner page:
```
<ons-page class="page page--MODIFIER-1" modifier="MODIFIER-1">
  <div class="page__background page--MODIFIER-1__background page--MODIFIER-2__background"> </div>
  <div class="page__content page--MODIFIER-1__content page--MODIFIER-2__content"></div>
  <div class="page__extra" style="z-index: 10001;"></div>
</ons-page>
```

MODIFIER-2 is inherited from the parent page.

__How to reproduce__
  

  
  [__Demo link__](https://tutorial.onsen.io/?issue)

  

- __HTML__

```html
<div id="app"></div>
```

- __JS__

```jsx
var MyPage = React.createClass({
  handleClick: function() {
    ons.notification.alert('Hello world!');
  },

  renderToolbar: function() {
    return (
      <Ons.Toolbar>
        <div className='center'>My app</div>
        <div className='right'>
        <Ons.ToolbarButton>
          <Ons.Icon icon='ion-navicon, material:md-menu'></Ons.Icon>
        </Ons.ToolbarButton>
        </div>
      </Ons.Toolbar>
    );
  },

  render: function() {
    return (
      <Ons.Page renderToolbar={this.renderToolbar} modifier="MODIFIER-2">
        
        <Ons.Page modifier="MODIFIER-1">
          <p style={{textAlign: 'center'}}>
            <Ons.Button onClick={this.handleClick}>
              Click me!
            </Ons.Button>
          </p>
        </Ons.Page>
        
        
      </Ons.Page>
    );
  }
});

ons.ready(function() {
  ReactDOM.render(<MyPage />, document.getElementById('app'));
});
```

PR #2240, opened 2017-11-01
fix(core, issue #2237): limited modifier scope

hugofelp:
This fixes the issue reported here - https://github.com/OnsenUI/OnsenUI/issues/2237 - where the modifier classes leak into child components of the same type.


frandiox:
@hugofelp Hi! Thanks a lot for the issue report and PR. I'll try to have a look at this next week when I have some time!

frandiox:
@hugofelp Thanks for the fix! I can't see any problem with this change :)

PR #2263, opened 2017-11-29
Weak map polyfill

frandiox:
@asial-matagawa @asialgearoid I think this should be enough :+1: 

Another possible solution would be importing `setup` before `ons` in `index.umd.js` in order to add all the polyfills. However, this is not possible for `index.esm.js` because `ons` is an external dependency and will be hoisted (`import` is always hoisted) when the user bundles his app. I think we should keep the same import order in both UMD and ESM to detect errors like this one.


asial-matagawa:
@frandiox 
Thank you! I think it is fine.

PR #2267, opened 2017-11-30
Fix tappable ons-list-item with tappable content

zabojad:
When we have a tappable ons-list-item with tappable content (buttons, inputs, ...), the ons-list-item "reacts" to `touchstart` and `touchend` events without the possibility to shortcut it. Even listening to thoses events with the `useCapure` flag set to `true` from the tappable content elements and calling `e.stopPropagation()` and `e.preventDefault()` won't help because `ons-list-item` is a web component and its internal event listeners seem to be called before any other...


zabojad:
PS: by the way, I've asked about this yesterday there: https://community.onsen.io/topic/1833/button-in-tappable-listitem-issue

zabojad:
@frandiox any thought on this one? :stuck_out_tongue:

frandiox:
@zabojad Hi, thanks for the PR! I was thinking about this after reading your question in the forum and came up with a possible solution. Your PR works in many cases but not in all of them because some components are made of plain `span` and `div`. Therefore, rather than hardcoding a list of elements, let's do it in two steps:

1. I think every `/^ons-/i.test(event.target.tagName)` can be ignored.
2. Any `event.target.hasAttribute('prevent-tap')` can also be ignored (attribute manually specified by the user).
3. If there is any element between `event.target` and `this` (the ons-list-item itself) in the DOM tree that fulfills any of the previous statements, it should also be ignored:

```javascript
util.findParent(
  event.target,
  parent => /^ons-/i.test(parent.tagName) || parent.hasAttribute('prevent-tap'),
  parent => parent === this
);
```

What do you think about that?

CC @asial-matagawa

zabojad:
@frandiox I like your proposition. I think that your point/step 2 alone would be enough actually, don't you think ? Checking only `parent.hasAttribute('prevent-tap')` would also save on perfs...

frandiox:
@zabojad It's enough but I guess most of the times the attribute will be added to `ons-button`, `ons-switch`, `ons-radio` or `ons-checkbox`, so checking the `tagName` can be convenient. A simple `event.target.tagName.indexOf('ONS-') === 0` would also do the trick and is a bit faster than a regex.

If we only want to check the attribute, the second argument for `util.findParent` can be a DOM query like `'[prevent-tap]'` instead of a function. Anyway, let's wait for more feedback.

asial-matagawa:
@frandiox 
I think all of the three steps are required too.
Adding the `prevent-tap` attribute to all `ons-*` elements is a bit tedious...

zabojad:
@frandiox OK then should I modify my PR to implement those 3 steps as described? Will that be OK to be merged then?

frandiox:
@zabojad Yes, please modify it :)
It would also be nice to have test cases for this in `ons-list-item.spec.js` :+1:

zabojad:
@frandiox done! Can you check and merge if it's OK?

frandiox:
@zabojad Thanks for the changes! I've modified it a little bit in #2280 . Already merged :)

zabojad:
@frandiox oups, sorry for the mistake... And thanks for merging !

PR #2277, opened 2017-12-08
Add popToTop() to ons-navigator and expose it to react binding

zabojad:
Hi !
This PR adds a `popToTop()` method to `ons-navigator` and expose it in the `Navigator` react binding.
Please review, comment and merge if OK.
Best.

### DIFF
--- a/bindings/react/src/components/Navigator.jsx
+++ b/bindings/react/src/components/Navigator.jsx
@@ -173,6 +173,35 @@ class Navigator extends BasicComponent {
   }
 
   /**
+   * @method popToTop
+   * @signature popToTop(options = {})
+   * @return {Promise}
+   *   [en] Promise which resolves to the revealed page.[/en]
+   *   [ja]明らかにしたページを解決するPromiseを返します。[/ja]
+   * @description
+   *   [en] Pops all pages out of the page stack except the top page[/en]
+   *   [ja][/ja]
+   */
+  popToTop(options = {}) {
+    if (this.isRunning()) {
+      return Promise.reject('Navigator is already running animation.');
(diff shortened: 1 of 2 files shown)


zabojad:
@frandiox this is to address https://community.onsen.io/topic/1820/ons-navigator-and-slide-animation-from-right-to-left/4 and in general add a feature common to most projects.

frandiox:
@zabojad Thanks again for another PR! I believe, however, this should be implemented by modifying the behavior of existing methods (through `options` object) instead of creating new methods. Especially considering this one is quite long and will need maintenance. Also, we use `top` nomenclature to refer to the exact opposite concept in `bringPageTop`, so it should be called something like `popHome` or something like that.

In any case, I've made a quick implementation of `resetToPage` with `options.pop` in https://github.com/OnsenUI/OnsenUI/commit/5855145dc93cf1a83a1202fee01255986ad39946 . Can you implement an option like this in React bindings' `resetPageStack` method? I think it should set the new stack + the old topPage, then call `popPage` method instead of `pushPage`.

zabojad:
@frandiox Thank you for your review! I've created a new PR implementing your suggestions: https://github.com/OnsenUI/OnsenUI/pull/2284.

By the way, who is the developper in charge of the react binding at onsen? I've some questions/suggestions regarding their implementation.

zabojad:
BTW, I guess this PR can be closed now...

frandiox:
@zabojad The original creators left long ago. React bindings are basically community driven + me reviewing and refactoring/implementing small features. I think somebody will take care of this from January/February, hopefully :)

PR #2281, opened 2017-12-09
Fix: Non-passive event listeners being used when they weren't necessary

### LINKED ISSUE #2249: Use passive event listeners for touch events
__Environment__ 

```
[Core]
    onsenui 2.6.1

[Framework]
    react 15.6.2

[Framework binding]
    react-onsenui 1.6.1

[Browser]
    Desktop - Chrome 62
```

__Encountered problem__
Chrome prints out a debug warning that the ListItem component uses a scroll blocking event:

> onsenui.js:3223 [Violation] Added non-passive event listener to a scroll-blocking 'touchstart' event. Consider marking event handler as 'passive' to make the page more responsive. See https://www.chromestatus.com/feature/5745543795965952

(You have to select 'Verbose' in the console to see them).

From what I can see in the code the event handlers never cancel the event so Onsen can easily use passive listeners. Maybe with a check if passive listeners are supported on the current browser.

I could submit a pull request, although it would be for 1.6 since I currently cannot upgrade to 1.7 on my project.

__How to reproduce__
Create a list with ListItems in it.

jmigual:
Fixes #2249

### DIFF
--- a/core/src/elements/ons-list-item.js
+++ b/core/src/elements/ons-list-item.js
@@ -214,10 +214,10 @@ export default class ListItemElement extends BaseElement {
   _setupListeners(add) {
     const action = (add ? 'add' : 'remove') + 'EventListener';
     this[action]('drag', this._onDrag);
-    this[action]('touchstart', this._onTouch);
+    this[action]('touchstart', this._onTouch, { passive: true });
     this[action]('mousedown', this._onTouch);
     this[action]('touchend', this._onRelease);
-    this[action]('touchmove', this._onRelease);
+    this[action]('touchmove', this._onRelease, { passive: true });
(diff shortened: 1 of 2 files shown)


frandiox:
@jmigual Hi! Thanks for the PR. This is something we want to add but I believe it's a bit more complex than this.

We still support some old browsers like Safari for iOS 8 and Android 4.4.4 and those do not support an object as the third parameter of `addEventListener`. Therefore, it will be considered as `true` boolean parameter. Also, the one in `ons-pull-hook` is specifically calling `preventDefault` so it cannot be passive.

We'll try to address this issue in a few weeks.

jmigual:
Oh I understand, thank you for explaining. Should we close de PR then?

PR #2284, opened 2017-12-11
React: Navigator resetPageStack pop option

zabojad:
Following discussions on #2277 with @frandiox, this PR adds a `pop` option to `Navigator.resetPageStack()`.

### DIFF
--- a/bindings/react/src/components/Navigator.jsx
+++ b/bindings/react/src/components/Navigator.jsx
@@ -203,6 +203,7 @@ class Navigator extends BasicComponent {
     }
 
     this.routesBeforePop = this.routes.slice();
+    this.routesAfterPop = this.routesBeforePop.slice(0, this.routesBeforePop.length - 1);
 
     const update = () => {
       return new Promise((resolve) => {
@@ -244,7 +245,7 @@ class Navigator extends BasicComponent {
 
     event.routes = {
       poppedRoute: this.routesBeforePop[this.routesBeforePop.length - 1],
-      routes: this.routesBeforePop.slice(0, this.routesBeforePop.length - 1)
+      routes: this.routesAfterPop
     };
 
     this.props.onPostPop(event);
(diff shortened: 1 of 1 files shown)


zabojad:
@frandiox can this be merged or is there anything wrong?

frandiox:
@zabojad Hey! I made a comment on a commit, didn't you see it?

zabojad:
Hey @frandiox ! On which commit have you made a comment ? I do not see it... Is it on 
a62f9f9 or a3cbc5a ?

frandiox:
@zabojad Hmm, maybe I'm the only one who can see it :sweat_smile: 

>@zabojad Doesn't this (`this.pages = routes.map(renderPage)`) need to be called for `options.pop` as well? I guess that's why you moved `var renderPage` to the top?

zabojad:
@frandiox it is one of the reasons I did ask you recently who is in charge of the React bindings in your team. Some of the OnSenUI React components sources I've crossed are strange (if not wrong) with things like calls to [`forceUpdate()` which is not recommended](https://reactjs.org/docs/react-component.html#forceupdate) (`setState()` should be used instead) or with variables that do not seem to be used at all. Here, in the Navigator component, the `this.pages` variable is anyway useless/never read because the [`render()` method will anyway regenerate the pages from the `routes` variable](https://github.com/OnsenUI/OnsenUI/blob/a62f9f9a7d87d6a70b658d3b7ad51f073da70096/bindings/react/src/components/Navigator.jsx#L325).

So I kept this `this.pages = routes.map(renderPage)` in `resetPageStack()` as it is done in the other methods as well but it is actually never used.

I think the React bindings might need some refactoring / rewriting to match both React programming good practices as well as OnSenUI components implementation...

For now, I can put back down this line which should not have gone up. We can discuss the general implementation of the react bindings in another thread... What do you think?

frandiox:
>I think the React bindings might need some refactoring / rewriting to match both React programming good practices as well as OnSenUI components implementation...
For now, I can put back down this line which should not have gone up. We can discuss the general implementation of the react bindings in another thread... What do you think?

@zabojad Yes, I agree, it needs refactoring. I already did [some](https://github.com/OnsenUI/OnsenUI/pull/2197) but not enough. Feel free to open a new issue/PR and suggest modifications :+1:

zabojad:
@friandox well the good thing here is that it can be done component by component so it's completely do-able...

I'll open new PR-s soon.

PR #2287, opened 2017-12-12
Completed build instructions when using some bindings

msaelices:
Otherwise the `gulp serve` command will fail if use some bindings on them


frandiox:
@msaelices Thanks!

PR #2298, opened 2018-01-04
Mithril bindings

FREEZX:
I've been working on developing OnsenUI bindings for mithril, as it's been my first choice when developing client apps for quite some time now. I have developed several apps using OnsenUI with mithril, but the integration i had was made in a hurry, so it wasn't pretty nor complete.

This is my attempt at creating proper mithril bindings. So far i have the navigator and the splitter properly working. I'll be actively working more to expand the features and capabilities, as i personally need it for my projects.

I'm making this PR for review purposes and feedback for now, until i have more things working.


frandiox:
@FREEZX Hi! Looks awesome! I'm not an expert in Mithril (actually I just learned it now in 20 min) but I'll try to give you some feedback.

Everything so far looks reasonable but I find the controllers (`NavigatorController` and `SplitterController`) a bit confusing for the end user. For example, the navigator has the `Navigator` component itself, a `pageStack` which seems to be read-only for the user (after creation) and then `NavigatorController` for passing a delegate to child pages. Perhaps I misunderstood something. Am I wrong?

As a suggestion, perhaps you can provide `Navigator.createPageStack([ initialPages ])`, which returns an array with overridden methods to allow calling `redraw` or `update` after push/pop/slice/etc. Some kind of reactivity. I guess this might not be very "Mithril-like", but it's just a suggestion for the API. Otherwise, perhaps the Navigator component itself could be passed down to the pages directly instead of a NavigatorController? I just feel there are too many entities, but maybe it's just my impression.

Also, do you think it's possible to plug Mithril's router with `ons-navigator`? I'm not sure if the router is configurable enough for that...

A bit unrelated to the code but it would be great if we can use Rollup in all the bindings instead of Webpack. This is not a requirement, just something nice to have.

If you continue with the bindings we can make you a member of the repo so you can freely contribute to both Mithril bindings and the core if necessary :+1:

FREEZX:
@frandiox Hi and thanks for the feedback!

I realise the controllers might be a bit confusing, but unfortunately i couldn't find a better way at the moment. I first tried adding functions in the classes, but i found it's not possible to keep references to the mithril's objects. Mithril didn't accept instantiated objects while keeping their properties, just classes, so this controller option seemed like the best alternative at the time of writing.

The page stack is not exactly read-only, but because of the same reason as above, i'm unable to keep references to the Navigator object, so i cannot easily access them if they are kept there. What i could do is keep all those properties and functions in the controller itself.

An alternative to remove entities out of the equation is to have a function on the "controller", returning it's own renderable component based on the controller's own data, which i think could work fairly well, and is quite mithril-like.

I think integrating `ons-navigator` with mithril's router would work just fine.
Using rollup won't be a problem, webpack is just there because when i started i just copied the vue folder.

I'd first like to get it to a stable enough state with more features before i become a member of the repo and contribute directly. I'll post when i get a new feature in.

emccorson:
I'm going back over this PR and I'm going to close it for now. At the moment, we would not add this to the core Onsen UI repository as Mithril is not a major framework. Instead, I think this would be better as a separate package (and maintained by @FREEZX or others, not the Onsen UI team).

@FREEZX If you are still working on these bindings and need help with them, please let us know and we'll do our best to assist. Thank you for your contribution to the Onsen UI bindings.

PR #2309, opened 2018-01-15
Add defaultValue props to react inputs

zabojad:
To be able to use seamlessly the onsen react inputs as [uncontrolled components](https://reactjs.org/docs/uncontrolled-components.html), we need them to support a [`defaultValue`](https://reactjs.org/docs/uncontrolled-components.html#default-values) property.

### DIFF
--- a/bindings/react/src/components/BaseInput.jsx
+++ b/bindings/react/src/components/BaseInput.jsx
@@ -25,8 +25,6 @@ class BaseInput extends BasicComponent {
 
     if (this.props.value !== undefined) {
       node.value = this.props.value;
-    } else if (this.props.defaultValue !== undefined) {
-      node.value = this.props.defaultValue;
     }
 
     this.EVENT_TYPES.forEach(eventType => node.addEventListener(eventType, this.onChange));
@@ -59,10 +57,15 @@ BaseInput.propTypes = {
   modifier: PropTypes.string,
   disabled: PropTypes.bool,
   onChange: PropTypes.func,
+  defaultValue: PropTypes.oneOfType([
+    PropTypes.string,
+    PropTypes.instanceOf(Date)
+  ]),
   value: PropTypes.oneOfType([
     PropTypes.string,
     PropTypes.instanceOf(Date)
   ]),
+  defaultChecked: PropTypes.bool,
   checked: PropTypes.bool,
   placeholder: PropTypes.string,
   type: PropTypes.string,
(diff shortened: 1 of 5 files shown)


frandiox:
@zabojad Hi! Thanks for the PR again. I have a doubt about this. Does `defaultValue` work as well as a prop? I thought it was only useful as an attribute on native input elements, like `<input defaultValue={...}>`.

Also, could you please add it to `Select`, `Checkbox` and `Radio` (`defaultChecked` for the last 2) as well for the sake of completeness? Thanks!

zabojad:
@frandiox yes you are right. I'm however facing the following issue: `defaultValue` on `ons.Input` seems to work already actually... However, `defaultChecked` does not work on `ons.Checkbox` and `ons.Radio` and I cannot figure out why... Do you have any clue on this one? 

I'll commit what I've done so far so that you see what I'm talking about...

zabojad:
@frandiox pushed. Can you have a look?

zabojad:
just edited my previous message...

zabojad:
@frandiox the problem is that it seems I need to propagate the `defaultValue` attribute into the core `ons-input` or `ons-radio`/`ons-checkbox` component... How to do that correctly?

zabojad:
@frandiox a little hint would be very appreciated here...

Here is what I've managed to do: I've passed `defaultValue` to the `ons-checkbox` component. There I've add this:
```
export default class BaseCheckboxElement extends BaseInputElement {

// ...
  _compile() {
    super._compile();

    this._setDefaultChecked();
  }

  _setDefaultChecked() {
    if (this.hasAttribute('defaultChecked')) {
      this._input.defaultChecked = this.getAttribute('defaultChecked');
    }
  }
// ...
```

It's odd because the input gets `checked` by default but its "style" is still unchecked...

Here is what I have at runtime:
screen shot 2018-03-01 at 17 22 34

Any idea why its style is still looking like it's unchecked?

zabojad:
@frandiox I'm not sure but I think it doesn't make sense to set a `defaultValue` attribute on `ons-...` internal native `input` as anyway `defaultValue` and `defaultChecked` are managed only by react in a react context and what happens inside the ons web component is out of reach for react...

I thus think that the right solution will be to set once `node.value = this.props.defautValue;`in the `componentDidMount()` method of the `BaseInput` react binding...

I'll issue another commit doing this...

frandiox:
@zabojad Hi! I think changing `node.value` and `node.checked` is the correct way, yes :+1: I'm not sure if there is another way to let React manage the input in this case. Just two things, I've seen a `var p = ...` that you probably forgot to remove and also a propType `defaultvalue` (with `v` lowercase). Could you change that? Thanks!

zabojad:
hey @frandiox! Oups, yes I'm gonna clean that. Things were getting a bit messy while debugging.

zabojad:
@frandiox done! Can it be merged already :)?

frandiox:
@zabojad Yep, thanks!

PR #2314, opened 2018-01-16
chore(fastclick): Use Onsen UI fork of FastClick to fix various issues

asialgearoid:


asialgearoid:
That is true, there is no global reference anymore. Though they should have no need to access it, because all you would do with `window.FastClick` is call `FastClick.attach`, which we do in `setup.js`. Any further interaction should be done with the returned instance, which we expose on `ons.fastclick`.

If you think there is a chance someone is using `window.FastClick` for some reason, we can add it to the `window` object after importing it in `setup.js`.

frandiox:
@asialgearoid That's true. I thought we were saying somewhere to run `ons.fastClick.destroy()` so they would need `window.FastClick` to attach it again. However, now that I think about it again, perhaps it was an issue related to `ons.GestureDetector` and not `ons.fastClick`. Nevermind, I think it's fine like this. Thanks!

PR #2315, opened 2018-01-16
Scroll/Swipe issues and passive event listeners

frandiox:


frandiox:
@asialgearoid Whenever you have time, could you please have a quick look at this? I've added the passive option to every event that needs it except when we actually need to prevent default behavior for fixing bugs on iOS. I've been checking iOS Safari 10 and I believe it works well. I think UIWebView and Safari <= 9 don't support passive listeners so it shouldn't change anything.

core/src/elements/ons-list-item.js
```diff
@@ -232,16 +236,9 @@ export default class ListItemElement extends BaseElement {
...
-      e => e.hasAttribute('prevent-tap') || /^ons-(?!col$|row$)/i.test(e.tagName),
-      e => e === this
-    ];
-  }
-
   _onTouch(e) {
-    if (this.tapped || !this._tapCheck()[1](e.target) &&
-      (this._tapCheck()[0](e.target) || util.findParent(e.target, ...this._tapCheck))
+    if (this.tapped || this !== e.target &&
+      (this._shouldIgnoreTap(e.target) || util.findParent(e.target, this._shouldIgnoreTap, p => p === this))
```
asialgearoid: It looks like adding `prevent-tap` on `ons-list-item` itself will have no effect, only its children. Is this correct?

On another note, I found this slightly confusing to read. Maybe add one extra set of brackets so it reads like this:

```
this.tapped || 
(this !== e.target && (this._shouldIgnoreTap(e.target) || util.findParent(e.target, this._shouldIgnoreTap, p => p === this))
```

core/src/elements/ons-pull-hook.js
```diff
@@ -173,9 +175,11 @@ export default class PullHookElement extends BaseElement {
     // Hack to make it work on Android 4.4 WebView and iOS UIWebView. Scrolls manually
```
asialgearoid: We are only applying this for iOS now, it seems, so it's worth updating this comment to reflect that

core/src/elements/ons-pull-hook.js
```diff
@@ -424,7 +436,8 @@ export default class PullHookElement extends BaseElement {
       this._gestureDetector = new GestureDetector(this._pageElement, {
         dragMinDistance: 1,
         dragDistanceCorrection: false,
-        dragLockToAxis: !this._dragLockDisabled
+        dragLockToAxis: !this._dragLockDisabled,
+        passive: !this._shouldFixScroll
```
asialgearoid: Can we set this to `passive: true` if we use the `overflow: hidden` scroll fix?

core/src/ons/util.js
```diff
@@ -20,11 +20,37 @@ import internal from './internal';
...
+  window.addEventListener('testPassive', null, opts);
+  window.removeEventListener('testPassive', null, opts);
+} catch (e) { null; }
+
+/**
+ * @param {Element} el Target
+ * @param {String} name Event name
+ * @param {Function} handler Event handler
+ * @param {Object} [opt] Event options (passive, capture...)
+ * @param {Boolean} [isGD] If comes from GestureDetector. Just for testing.
```
asialgearoid: Do we need to keep this param?

frandiox: @asialgearoid It's internal so I guess it's not an issue :sweat_smile: . It is just for debugging in the future.

core/src/elements/ons-pull-hook.js
frandiox: @asialgearoid I don't think so, the pull-hook flickering issue is not related to the `overflow: hidden` thing on iOS

core/src/elements/ons-pull-hook.js
frandiox: This one was also necessary for Android, it's a different issue from the iOS page scroll flickering. Actually I could not reproduce it in my Android 7 but I remember seeing the issue in older versions some time ago, so it's safer to keep it.

core/src/elements/ons-list-item.js
frandiox: That's correct, instead of `prevent-tap` they should just remove `tappable` from `ons-list-item`. I'll try to make that more readable, thanks!

core/src/elements/ons-pull-hook.js
asialgearoid: I've just realised I mis-read the line `this._shouldFixScroll = util.globals.actualMobileOS !== 'other';` as `actualMobileIOS`. Ignore this comment, and also the next one!

PR #2317, opened 2018-01-17
fix(platform): Add landscape check for iPhone X. Closes #2279.

### LINKED ISSUE #2279: Landscape iPhone X
Onsen fails to detect the iPhone X platform if the app is started in Landscape.

The ons.platform.isIPhoneX() function should include both portrait and landscape options ((h 812 && w 375) || (w 812 && h 375)).

Note; I was following [this](https://onsen.io/v2/guide/iphonex.html) and applying the setAttribute during the initial boot of my app. At the moment if the device is in landscape it does not add the setAttribute.

asialgearoid:
Though this check isn't strictly necessary for on-device use, it is helpful for in-browser development. See #2279 for discussion.


frandiox:
@asialgearoid Yep, I agree we should support dev env as well, thanks!

PR #2322, opened 2018-01-19
iOS scroll fix

frandiox:
@asialgearoid This simple fix seems to be enough for all the crazy iOS scroll issues. At least it works well in UIWebView on iOS 11. I could reproduce all the issues by removing the previous "half fix" and fix them all with ef537ad
What do you think about it? (only the last commit, the rest are part of #2315)

cc @Ralpharoo


asialgearoid:
@frandiox I did consider this option, but had problems with it in the past. Specifically, setting `overflow: hidden` would reset `scrollTop` to `0`. So, if the user had scrolled down the screen, and then opened an overlay, it would jump back up to the top of the page. In the end, this was a much more frustrating behaviour than the overlay scrolling the element behind it. Did you test that scenario on device?

frandiox:
@asialgearoid Actually I tested that on both iOS UIWebView and Safari, and it was working well. I thought it would scroll to the top but it did not. Maybe we should test it further...

asialgearoid:
I've done some testing and confirmed that `overflow: hidden` does sometimes reset the scroll bar to the top on iOS. It does not appear to happen when we set it on `.page__content`, but honestly I have no idea why that is.

frandiox:
Continues in #2328

PR #2330, opened 2018-01-26
Place the cursor at the end of the text of the prompt text input.

laelbe:


frandiox:
@laelbe I guess this is for the `options.defaultValue`. Makes sense, thanks!

PR #2338, opened 2018-01-30
fix(react): support <Input readonly=false />

### LINKED ISSUE #2323: Missing readOnly definition in ReactJS Input
Platform ReactJS

The Input.propTypes is missing the definition of the readOnly attribute as a boolean. As a result, readOnly={false} still generates the attribute in the markup and makes the Input readOnly. Should be defined similarly to disabled.

sbusch:
* add proptype to <Input /> for readonly prop to support `... readonly={false}`, which closes #2323
* add example for readonly=true

(a commit contained in this PR wrongly states "... readonly=true", this should be "false")

### DIFF
--- a/bindings/react/examples/components/Input.js
+++ b/bindings/react/examples/components/Input.js
@@ -55,7 +55,9 @@ export default class extends React.Component {
         <p>
         Please enter a text
         </p>
-          <Input disabled={false} value={this.state.text} float onChange={event => this.setState({text: event.target.value}) } modifier='material' placeholder='Username'></Input>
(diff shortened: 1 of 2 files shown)


frandiox:
@sbusch Looks good, thanks!

frandiox:
@sbusch I was thinking that, while this works, wouldn't it be better the camel case version `readOnly`? https://reactjs.org/docs/dom-elements.html#all-supported-html-attributes

sbusch:
Oh yes, wasn't aware of that! Thanks.

Fixed in PR #2339

PR #2339, opened 2018-01-31
fix(react): change case of "readonly" prop to "readOnly"

sbusch:
to be in conformance to React, see <https://reactjs.org/docs/dom-elements.html#all-supported-html-attributes>

Follow-up to PR #2338


sbusch:
Do not merge yet, example needs to be changes aswell. Commit follows...

sbusch:
Ok, PR is ready

frandiox:
@sbusch Thanks!

PR #2347, opened 2018-02-07
Update changed className correctly

### LINKED ISSUE #2158: Unsynchronized `class` and `className` in React binding of OnsenUI
__Environment__ 

```
[Core]
    onsenui 2.1.0

[Framework]
    react 15.4.2

[Framework binding]
    react-onsenui 1.1.0

[Additional libraries]
    none

[Platform]
    Desktop - macOS Sierra

[Browser]
    Desktop - Chrome/60.0

[UserAgent]
    Mozilla/5.0 (Macintosh; Intel Mac OS X 10_12_6)
    AppleWebKit/537.36 (KHTML, like Gecko)
    Chrome/60.0.3112.101 Safari/537.36
```

__Encountered problem__
When adding a dynamic `className` property to a `react-onsenui` component (specifically, `ListItem`), when the property changes, the `class` item on the DOM element itself doesn't change properly. For instance, when changing a component's `className` from `box black` to `box`, the `class` property on the DOM element won't remove the `black` class from itself.

__How to reproduce__
Change a `ListItem` component's class from `box black` to `box` -- the `class` property on the DOM element won't remove the `box` class from itself.

burka:
This fixes #2158.

See https://codepen.io/anon/pen/OQRKeW?editors=1111 to reproduce the error.

### DIFF
--- a/bindings/react/src/components/BasicComponent.jsx
+++ b/bindings/react/src/components/BasicComponent.jsx
@@ -21,9 +21,9 @@ class BasicComponent extends React.Component {
         node.className = node.className.replace(this.lastClass, ' ');
       }
 
-      this.lastClass = ' ' + this.props.className.trim();
+      this.lastClass = this.props.className.trim();
 
-      node.className = node.className.trim() + this.lastClass;
+      node.className = node.className.trim() + ' ' + this.lastClass;
     }
 
     if (!ons) {


frandiox:
@burka Thanks for the fix! I actually think updating classes is not necessary anymore. In recent versions, the core elements restore their own class when needed so I think we could remove this logic from the React side...

PR #2348, opened 2018-02-08
popover position should be related to page.

laelbe:
boxed

There is a bug with boxed sized application.


frandiox:
@laelbe Hi, thanks for the PR! Could you please submit a small example where the issue can be seen? Also, I think we cannot use `this.closest` since it is not supported on iOS 8.

laelbe:
Here is the example. 
[Codepen Example](https://codepen.io/anon/pen/YexvGa)

Go "Dialogs" > "More"

laelbe:
bug1
And, also, there is another bug with ons-popover with long element.

asialgearoid:
Thanks for the PR, @laelbe. This doesn't entirely fix the issue as I think it would break if the popover is placed in the body. I have fixed it in a separate PR (https://github.com/OnsenUI/OnsenUI/pull/2392). Your PR was very useful for helping with the fix. Thank you very much!

PR #2355, opened 2018-02-21
Fixes "Cannot read property 'classList' of null

renanbandeira:
I'm facing a problem that when I have a carousel on navigationStack and I need to remove this page, the app fires this error on Console and the navigation. As far as I understood, Onsenui tries to toggle the active carousel item through css, but the component has been already unmounted.

### DIFF
--- a/core/src/ons/internal/swiper.js
+++ b/core/src/ons/internal/swiper.js
@@ -143,7 +143,10 @@ export default class Swiper {
   }
 
   _setSwiping(toggle) {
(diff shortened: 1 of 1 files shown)


frandiox:
@renanbandeira Hi, thanks for the PR! Could you please check what is calling `_setSwiping` in your situation (error stack trace)? Perhaps there is a listener that is not being removed or something similar that we should fix instead of just adding an `if` statement.

Issue #2335, opened 2018-01-29
State Change In React-OnsenUI Tabbar


johnkmzhou:
__Environment__

  
```
[Dependencies]
    "onsenui": "^2.9.1",
    "react": "^16.2.0",
    "react-dom": "^16.2.0",
    "react-onsenui": "^1.10.0",
    "react-redux": "^5.0.6",
    "react-scripts": "1.1.0",
    "redux": "^3.7.2"
```

__Encountered problem__
When performing state changes on the renderTabs prop of Tabbar only the tab will re-render but not the corresponding page.

  
  [__Demo link__](https://github.com/johnkmzhou/onsenui-tabbar)

  
- __HTML__

```html
```

- __JS__

```jsx
import React from 'react';
import { Page, Tabbar, Tab } from 'react-onsenui';
import { compose } from 'redux';
import { connect } from 'react-redux';

import HomePage from './HomePage';
import LoginPage from './LoginPage';
import SettingsPage from './SettingsPage';

const AppTabbar = props => {
  const login = navigator => [
    {
      content: <HomePage key="home" navigator={props.navigator} />,
      tab: <Tab key="home" label="Home" icon="md-home" />
    },
    {
      content: <LoginPage key="login" navigator={props.navigator} />,
      tab: <Tab key="login" label="Login" icon="md-account-o" />
    }
  ];

  const settings = navigator => [
    {
      content: <HomePage key="home" navigator={props.navigator} />,
      tab: <Tab key="home" label="Home" icon="md-home" />
    },
    {
      content: <SettingsPage key="settings" navigator={props.navigator} />,
      tab: <Tab key="settings" label="Settings" icon="md-settings" />
    }
  ];

  return (
    <Page>
      <Tabbar position='bottom' index={0} renderTabs={props.loginPage ? login : settings} />
    </Page>
  );
};

export default compose(
  connect(({ tabReducer: loginPage }) => (loginPage))
)(AppTabbar);
```

snazzypants:
Bump...this bug makes it basically impossible to share state between tabs.

chronweigle:
I wrapped Tabbar to explicitly pass in an index prop, fixing this issue for me.

https://gist.github.com/chronweigle/c7bf3b5a7ad21dbbb5110236439ad4f6

PR #2367, opened 2018-03-04
React Tabbar, update page when not passing explicit prop in.

chronweigle:
This fixed issue #2335 for me.  I've only tested by wrapping the component, haven't had a chance to rebuild and test this patch to the library yet.

Wrapper for component:
https://gist.github.com/chronweigle/c7bf3b5a7ad21dbbb5110236439ad4f6

### DIFF
--- a/bindings/react/src/components/Tabbar.jsx
+++ b/bindings/react/src/components/Tabbar.jsx
@@ -12,14 +12,14 @@ import Util from './Util.js';
  * To define the tabs and the content the property renderTabs need to be implemented, that returns an array of tabs and their content. See the example for specifics. [/en]* [ja][/ja]
  * @example
 
-  <Page>
-    <Tabbar
-      onPreChange={({index}) => this.setState(index)}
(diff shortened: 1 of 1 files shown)


asialgearoid:
@chronweigle I'm not sure why this PR was not responded to originally, sorry for the delay.

I've tested it out, and it doesn't appear to actually fix the issue described in #2335. I can imagine this might fix issues where you were dynamically setting the active tab. However it does not seem to fix the issue where changing the actual tabs does not update the pages which the tabs relate to.

chronweigle:
I mentioned I hadn't tested it. Looking at this now I see it shouldn't work. I missed hooking into the onPostChange callback as I did in the wrapper in the gist.

I've opened a new PR and provided an explanation. I spent more time providing an explanation of the problem than the solution since I don't know if this is the ideal solution. A proper understanding of the problem should lead you to a solution.

Thank you

https://github.com/OnsenUI/OnsenUI/pull/2508

PR #2384, opened 2018-03-31
fix(chore): Add React 16 as peer dependency

### LINKED ISSUE #2257: Warning about react missing when using react 16?
I get:
```
npm WARN react-onsenui@1.9.0 requires a peer of react@^15.5.3 but none was installed.
npm WARN react-onsenui@1.9.0 requires a peer of react-dom@^15.5.3 but none was installed.
```

While I have react 16 installed:
```
  "dependencies": {
    "onsenui": "^2.8.1",
    "react": "^16.1.1",
    "react-dom": "^16.1.1",
    "react-onsenui": "^1.9.0",
```

sbaechler:
Fixes #2257

### DIFF
--- a/bindings/react/package.json
+++ b/bindings/react/package.json
@@ -91,8 +91,8 @@
   },
   "peerDependencies": {
     "onsenui": "~2.9.0",
-    "react": "^15.5.3",
-    "react-dom": "^15.5.3"
+    "react": "^15.5.3 || ^16.0.0",
+    "react-dom": "^15.5.3 || ^16.0.0"
   },
   "dependencies": {
     "prop-types": "^15.6.0"


asialgearoid:
@sbaechler I don't think we need to have both 15 and 16 in the version. Just 16 should be fine. I haven't tested with React 16 yet, but I expect there shouldn't be any problems.

sbaechler:
Since they are just peer dependencies, I would keep 15 until the next major release of OnsenUI. Otherwise all the people that still use React 15 are going to have the same annoying message that the React 16 users have now.

asialgearoid:
We've ended up adding just `^16.0.0` as a peer dependency, so I will close this PR.

PR #2400, opened 2018-04-16
Increase font size of text-input in order to avoid iOS zoom issues

b44rd:
Hi, just thought I could contribute with fixing a little annoying issue. 

When building Onsen apps, inputs has by default got a 15 pixel font size, which makes iOS/iPhone zoom unintended when interacting with the element. Increasing font-size to 16px seem to be the only good fix for this. So thought it could be nice to suggest this beeing the new default. 

Thanks for building an amazing framework, by the way. I´m really enjoying it.

### DIFF
--- a/css-components/src/components/text-input.css
+++ b/css-components/src/components/text-input.css
@@ -1,6 +1,6 @@
 
 :root {
-  --text-input-font-size: 15px;
+  --text-input-font-size: 16px;
   --text-input-height: 31px;
   --text-input-border-color: var(--input-border-color);
   --material-text-input-font-size: 16px;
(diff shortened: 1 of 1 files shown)


b44rd:
Hi, @asialgearoid, 

If you set user-scalable=no, you instruct the browser to disallow zoom completely. Though, later versions of iOS no longer care if user-scalable or not, which can be a bit problematic. 

Someone described parts of iOS and zooming very well in this (a bit old and partially outdated) blog post;
http://blog.osmosys.asia/2017/01/05/prevent-ios-from-zooming-in-on-input-fields/

To me a good solution then seem to be defaulting to 16 pixels in inputs?

Issue #2261, opened 2017-11-26
Reactive Forms NOT WORKING ON ANGULAR 2+


SebasG22:
## Update: HAVE PASSED 5 MONTHS WITHOUT ANSWER 

**Create at:** 26/11/2017
**Update at:** 17/05/2018

Hello everyone,

I started using OnsenUI finding that it is not possible to perform reactive forms that by default are the new feature of the new versions of angular (Angular + 2). This is a big problem since most applications use Forms for different functionalities. Currently it is possible to make "Template Driven Forms" but these involve much more code and process when making validations that with less code can be done with "Reactive Forms".

[Core]
    onsenui 2.8.2

[Framework]
    angular5 5.0.1

__Encountered problem__

Angular Compiler says: 
`LoginEmailComponent.html:3 ERROR Error: No value accessor for form control with name: 'email
![captura de pantalla 2017-11-26 a la s 7 37 06 a m](https://user-images.githubusercontent.com/17608169/33240039-13859e8a-d27d-11e7-8ffb-b5786bd599c5.png)`
```js
    at _throwError (forms.js:2385)
    at setUpControl (forms.js:2257)
    at FormGroupDirective.addControl (forms.js:6606)
    at FormControlName._setUpControl (forms.js:7256)
    at FormControlName.ngOnChanges (forms.js:7169)
    at checkAndUpdateDirectiveInline (core.js:12092)
    at checkAndUpdateNodeInline (core.js:13598)
    at checkAndUpdateNode (core.js:13541)
    at debugCheckAndUpdateNode (core.js:14413)
    at debugCheckDirectivesFn (core.js:14354)
View_LoginEmailComponent_0 @ LoginEmailComponent.html:3
```

__How to reproduce__

1. Create a component that contains a basic reactive form 

login-email.component.ts
```js
`this.form = new FormGroup({
      email: new FormControl('', Validators.required && Validators.email),
      password: new FormControl('', Validators.required)
    });
```
login-email.component.html
```html
<div style="padding-top: 10px">
      <form [formGroup]='form' (ngSubmit)="onSubmit(form)">
          <ons-input float placeholder="email" type="email" [formControlName]='email'></ons-input>
          <ons-input float placeholder="password" type="password" formControlName='password'></ons-input>
          <div style="text-align: center; margin: 10px;">
              <ons-button type="submit"> Sign in </ons-button>
            </div>
        </form>
</div>
```

See also :
[...]

PR #2401, opened 2018-04-18
Add ControlValueAccessor to ons-input

puku0x:
As mentioned in #2261, the current `ons-input` does not support Reactive Form.
This PR proposes how to implement it.

We can use `ons-input` directive as following with help of `ControlValueAccessor `.
```html
<form [formGroup]="form">
  <ons-input formControlName="value"></ons-input>
</form>
```
We can use `ngModel` too.
```html
<ons-input [(ngModel)]="value"></ons-input>
```

How do you like it ?

### DIFF
--- a/bindings/angular2/src/directives/ons-input.ts
+++ b/bindings/angular2/src/directives/ons-input.ts
@@ -11,11 +11,7 @@ import {
   SimpleChange,
   forwardRef
 } from '@angular/core';
-import { 
-  ControlValueAccessor,
-  FormControl,
-  NG_VALUE_ACCESSOR
-} from '@angular/forms';
+import { ControlValueAccessor, NG_VALUE_ACCESSOR } from '@angular/forms';
 
 /**
  * @element ons-input
@@ -40,6 +36,7 @@ import {
 export class OnsInput implements OnChanges, OnDestroy, ControlValueAccessor {
   private _element: any;
   private _boundOnChange: Function;
+  private _propagateChange = (_: any) => { };
 
   /**
    * @input value
@@ -66,25 +63,9 @@ export class OnsInput implements OnChanges, OnDestroy, ControlValueAccessor {
     this._element.addEventListener('input', this._boundOnChange);
   }
 
-
-  private propagateChange = (_: any) => { };
-
-  public writeValue(obj: any) {
-   if (obj) {
-     this._element.value = obj;
-   }
-  }
-
-  public registerOnChange(fn: any) {
-    this.propagateChange = fn;
-  }
-
-  public registerOnTouched() { }
-
-
   _onChange(event: any) {
     this._valueChange.emit(this._element.value);
-    this.propagateChange(this._element.value);
+    this._propagateChange(this._element.value);
   }
 
   ngOnChanges(changeRecord: {[key: string]: SimpleChange;}) {
@@ -107,4 +88,16 @@ export class OnsInput implements OnChanges, OnDestroy, ControlValueAccessor {
 
     this._element = null;
   }
+
+  writeValue(obj: any) {
+    if (obj) {
+      this._element.value = obj;
+    }
+  }
+ 
+  registerOnChange(fn: any) {
+     this._propagateChange = fn;
+  }
+ 
+  registerOnTouched() { }
 }
--- a/bindings/angular2/src/directives/ons-input.ts
+++ b/bindings/angular2/src/directives/ons-input.ts
@@ -5,6 +5,7 @@ import {
   ElementRef,
   Input,
   Output,
+  HostListener,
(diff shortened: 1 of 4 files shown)


asialgearoid:
@puku0x This is great, it works perfectly. I will merge it and add an example with some unit tests. Thanks for continued input, it is really appreciated.

asialgearoid:
@puku0x I just realised that if we are to add support for reactive forms, we will need it for all types, such as checkbox, radio and select. What do you think?

puku0x:
@asialgearoid Yes, I think it is needed that `ngx-onsenui`'s form directives should support reactive forms for all input types.

I noticed that some of them already have `ControlValueAccessor`. The remaining directives need to implement`ControlValueAccessor`.

- [x] ons-checkbox
- [ ] ons-input
- [x] ons-radio
- [ ] ons-range
- [ ] ons-search-input
- [x] ons-select
- [ ] ons-switch

asialgearoid:
@puku0x Thanks for checking. I will hopefully get a chance to take a look at these some time next week. If you have time to update your pull request for any of those directives, that would be amazing.

asialgearoid:
@puku0x This is great, thank you very much! I will take a look at this soon.

bindings/angular2/src/directives/ons-input.ts
```diff
@@ -77,4 +88,14 @@ export class OnsInput implements OnChanges, OnDestroy {
...
+
+  writeValue(obj: any) {
+    this._element.value = obj;
+  }
+ 
+  registerOnChange(fn: any) {
+     this._propagateChange = fn;
+  }
+ 
+  registerOnTouched() { }
```
asialgearoid: Is this empty `registerOnTouched` function actually required?

puku0x: Yes, it is required by `ControlValueAccessor ` interface.
https://angular.io/api/forms/ControlValueAccessor

asialgearoid: Thanks for the link to the docs. If we do not call the argument `fn` as described in the docs, does Angular still mark the fields as touched correctly?

puku0x: Unfortunately no.
I tried two examples but `touched` property didn't change to `true`.

```html
<form [formGroup]="form">
  <ons-input formControlName="value"></ons-input>
</form>
{{form.controls['value']?.touched}}  <!-- false -->
```

```html
<form #f="ngForm">
  <ons-input name="value" [(ngModel)]="value"></ons-input>
</form>
{{f.form.touched}}  <!-- false -->
```

asialgearoid: If we completely remove `registerOnTouched`, does it throw an error? My expectation would be that if it doesn't exist, Angular would use the default behaviour.

If removing `registerOnTouched` doesn't work, perhaps we can change it to:
```
registerOnTouched(fn: any) {
  fn();
}
```

puku0x: Removing `registerOnTouched` causes compile error. This is because of TypeScript specification.

I think this may be better.

```ts
private _propagateTouched = () => {};

@HostListener('blur')
_onBlur() {
  this._propagateTouched();
}

registerOnTouched(fn: any) {
  this. _propagateTouched = fn;
}
```

asialgearoid: Ah yes, of course. My suggestion would not work, because it was called in the registration handler. This looks like a good solution to me. :+1:

puku0x:
Based on above conversation, I added handler related to `registerOnTouched`.

emccorson:
Everything looks good. I added some unit tests and merged. Thanks for the PR!

puku0x:
Thank you @emccorson and @asialgearoid !

asialgearoid:
That would be fantastic. Thank you for your continued input, @puku0x, it is greatly appreciated.

SebasG22:
Hi @puku0x @asialgearoid , Thanks for add this feature.

I have tried to use natively reactive forms with ngx-onsenUI but I see that they have not yet made the release.

[The latest version released at npm is 4.1.0](https://www.npmjs.com/package/ngx-onsenui)  

My question is when will they launch this vital feature?

**Note**: I tried to use the [development version](https://onsenui.github.io/latest-build) but, I could never install the dev version. If you have any guidelines for using the development version locally, it would be very useful.

Thanks

emccorson:
@SebasG22 We are hoping to release reactive forms support with the next version of Onsen UI around the end of this month.

As for the development version, I got it working by running `npm link` from the unzipped Onsen directory, then `npm link onsenui` from my project directory. Hope this helps.

PR #2408, opened 2018-04-29
Remove dotted border on buttons in Firefox

DASPRiD:
This PR will remove the border which Firefox inserts on all focused buttons. The normal :focus { outline: 0; } does actually not do anything here.

### DIFF
--- a/css-components/src/components/button.css
+++ b/css-components/src/components/button.css
@@ -50,6 +50,10 @@
     outline: 0;
   }
 
+  --button::-moz-focus-inner {
+    border: 0;
+  }
+
   --button--active: {
     background-color: var(--button-background-color);
     transition: none;


asialgearoid:
Thanks for the PR, @DASPRiD, this is a good idea. I think it would be better to put it next to the existing `outline: 0` right here: https://github.com/OnsenUI/OnsenUI/blob/e4bbee238e801d5d2ec93cb653a4a8ea01211f95/css-components/src/components/button.css#L50

Could you change your code to put it there instead?

DASPRiD:
Sorry for the long wait, I've updated the code accordingly.

asialgearoid:
Great, thanks for the PR! :+1:

Issue #2420, opened 2018-05-09
Modal containing Toolbar doesn't respect iPhone X safe areas


karlvr:
__Environment__

Onsen UI Version:

- 2.9.2

Framework:

- N/A but React 16.3

Framework binding:

- N/A but reoct-onsenui 1.10.0

Additional libraries:

- N/A

Platform:

- iOS 11.3.1 Safari on iPhone X

__Encountered problem__
An `<ons-modal>` occupies the full screen, but the `iphonex-support/combination.css` removes the safe area margins from an `<ons-toolbar>` inside the modal, thereby removing the safe area margins and causing the toolbar in the modal to go under the dreaded notch.

__How to reproduce__

Below is a playground that would demonstrate the issue if run on an iPhone X.

  
  [__Demo link__](https://onsen.io/playground/?issue)

  

- __HTML__

```html
<div id="app"></div>
```

- __JS__

```jsx
var MyPage = React.createClass({
  
  handleClick: function() {
    this.setState({showModal: true});
  },

  renderToolbar: function() {
    return (
      <Ons.Toolbar>
        <div className='center'>My app</div>
        <div className='right'>
        <Ons.ToolbarButton>
          <Ons.Icon icon='ion-navicon, material:md-menu'></Ons.Icon>
        </Ons.ToolbarButton>
        </div>
      </Ons.Toolbar>
    );
  },

  renderModalToolbar: function() {
    return (
      <Ons.Toolbar>
        <div className='center'>My modal</div>
      </Ons.Toolbar>
    );
  },

  render: function() {
    return (
      <Ons.Page renderToolbar={this.renderToolbar}>
        <p style={{textAlign: 'center'}}>
          <Ons.Button onClick={this.handleClick}>
            Click me!
          </Ons.Button>
        </p>
        <Ons.Modal isOpen={this.state && this.state.showModal} animation="lift">
          <Ons.Page renderToolbar={this.renderModalToolbar}>
            <p>Modal here</p>
          </Ons.Page>
        </Ons.Modal>
      </Ons.Page>
    );
  }
});

document.documentElement.setAttribute('onsflag-iphonex-portrait', '');
document.documentElement.setAttribute('onsflag-iphonex-landscape', '');

ons.ready(function() {
  ReactDOM.render(<MyPage />, document.getElementById('app'));
});


```

PR #2421, opened 2018-05-09
Respect safe area margins inside Modal on iPhone X

karlvr:
This PR is to address issue #2420


asialgearoid:
Looks good, thanks for the fix! 👍

PR #2424, opened 2018-05-10
RouterNavigator: Fix double-mount

### LINKED ISSUE #2423: RouterNavigator double-mounts route components
__Environment__

Onsen UI Version:

- 2.9.2

Framework:

- React 16.3.2

Framework binding:

- react-onsen 1.10.0

Platform:

- macOS Chrome, iOS 11.3.1 Safari

__Encountered problem__

`RouterNavigator` creates new page components in such a way that components are mounted and rendered twice. This is because of the `render` method in `RouterNavigator`, which renders the `renderPage` of each route in its config, _and_ the currently animating page separately. During the animation the page is actually rendered twice, and then back to once, and so we get two components. This can result in a flash if you use components that use canvas etc.

__How to reproduce__

My package at https://github.com/karlvr/onsenui-react-redux-navigator has an example that demonstrates this issue. The `MyPage` class is mounted twice each time you push a new page on.

I have got a PR ready to address this.

karlvr:
Fixes #2423

### DIFF
--- a/bindings/react/src/components/RouterNavigator.jsx
+++ b/bindings/react/src/components/RouterNavigator.jsx
@@ -241,13 +241,21 @@ class RouterNavigator extends BasicComponent {
     this.cancelUpdate = true;
   }
 
+  componentDidUpdate() {
+    /* When the component updates we now have the page we're pushing in our routeConfig, so we no longer need to render it specially */
+    this.page = null;
+  }
+
   render() {
     const attrs = Util.getAttrs(this);
 
+    /* Gather pages to render and the animating page in one array so React reuses components. */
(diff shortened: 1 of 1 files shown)


asialgearoid:
Thanks for your PRs, @karlvr. I haven't spent much time yet working with the React bindings, so your contribution is much appreciated! It may take me a little while to test, but it looks good at first glance.

asialgearoid:
@karlvr Sorry it's taken so long to merge this. I've finally had some time to spend on the React bindings. Thanks for the PR!

karlvr:
@asialgearoid no problem at all, thank you!

PR #2425, opened 2018-05-10
chore(ngx-onsenui): Support Angular 6.0.0

puku0x:
Current `ngx-onsenui` supports Angular 4.x but the latest version is 6.0.0.
This PR updates`ngx-onsenui`'s dependencies such as `@angular/core` and `rxjs`.


asialgearoid:
@puku0x Thanks for the PR! It looks like this is just a matter of updating the dependencies and fixing some paths for the examples. Did you come across any other issues?

puku0x:
@asialgearoid 
No I didn't. I think there are no any other problems for now 🙂

asialgearoid:
OK great! I'll have a quick look at this on my machine and then merge it :+1:

asialgearoid:
@puku0x There are quite a few failing tests -- take a look at [the latest build result on CircleCI](https://circleci.com/gh/OnsenUI/OnsenUI/5642). If you run the tests locally (`cd bindings/angular2 && npm run test`), you should see the same results.

Unfortunately I think there are quite a few things that have broken as a result of this upgrade to Angular 6.

puku0x:
Oops, there is a problem in `compoent-loader.ts`.
`_rootComponents` property has omitted from `ApplicationRef` since Angular v5.
I'll update my PR soon.

asialgearoid:
@puku0x Thanks for the quick fix!

Upgrading CircleCI to Node v8 is fine. You can change it yourself in this file:
https://github.com/OnsenUI/OnsenUI/blob/master/.circleci/config.yml#L6

Change the image to `circleci/node:8-browsers`, and that should fix it.

asialgearoid:
It looks like it's failing now due to an issue with Sauce Labs. I don't think it has anything to do with your changes. I'll investigate now.

asialgearoid:
@puku0x It turns out that when CircleCI builds from pull requests, it does not include the private keys required for Sauce Labs. This is because someone could submit a PR which would log those publicly. I think this must be why CircleCI was originally disabled for PRs on here, because the Sauce tests will fail.

However, I ran the tests manually and it looks good. Thank you!

PR #2436, opened 2018-05-28
fix(ons-toast): Fixed layout of material design toast

MissMonacoin:
Fixed layout of ons-toast on Android.
Before fixing:

### DIFF
--- a/css-components/src/components/toast.css
+++ b/css-components/src/components/toast.css
@@ -74,7 +74,7 @@
   margin: 0;
   background-color: var(--material-toast-background-color);
   border-radius: 0;
-  padding: 24px;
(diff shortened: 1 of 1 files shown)


asialgearoid:
@MissMonacoin Thanks for the PR! We recently changed ons-toast to support multiple lines of text. It seems that it broke the padding of Material toasts. 

However, the vertical padding needs to be `16px` rather than `24px`. So, I think it should be:
```
padding: 16px 24px;
```

If you change your PR to that, then it is good to merge 👍

MissMonacoin:
Changed.

asialgearoid:
Great! ありがとうございました。

Issue #2475, opened 2018-07-11
ons.platform.isAndroid and ons.platform.isIOS take into account selected platform


emccorson:
Onsen UI Version: 2.10.3

The functions `ons.platform.isAndroid` and `ons.platform.isIOS` take into account any values set using `ons.platform.select`.

For example, if you are using iOS but have set the platform to `android`, `ons.platform.isAndroid` will return `true`.

```javascript
// on iOS
ons.platform.isAndroid(); //false
ons.platform.select('android');
ons.platform.isAndroid(); //true
```

This seems like unexpected behaviour to me. Maybe we could have separate functions like `ons.platform.isReallyAndroid` and `ons.platform.isReallyIOS`, or change the behaviour of the existing functions to ignore selected platform.

Also the documentation at https://onsen.io/v2/api/js/ons.platform.html#methods-summary should mention that the selected platform is not ignored.

asialgearoid:
It's hard to say what this behaviour should be. When the user specifically selects a platform, they may expect it to return true in this case. Perhaps we could offer a way of toggling the boolean `_ignorePlatformSelect` in `platform.js`, which is currently what controls whether these functions factor in the user's `select` setting. This would allow us to keep the current behaviour, but give the option for an override.

An update to the documentation is probably the most important part of this though, to make it clear to users exactly how this works.

PR #2477, opened 2018-07-12
feat(platform): can ignore selected platform when checking the platform

emccorson:
Enhancement based on #2475.

I've done this by adding `ignoreSelectedPlatform` argument to the relevant functions, with default of `false` so it doesn't change existing behaviour.

If there is a better way of doing this, let me know and I'll change it.

### DIFF
--- a/core/src/ons/platform.js
+++ b/core/src/ons/platform.js
@@ -138,17 +138,17 @@ class Platform {
   //----------------
   /**
    * @method isIOS
-   * @signature isIOS([ignoreSelectedPlatform])
-   * @param {Boolean} ignoreSelectedPlatform
+   * @signature isIOS([forceActualPlatform])
+   * @param {Boolean} forceActualPlatform
    *   [en]If true, selected platform is ignored and the actual platform is returned.[/en]
    *   [ja][/ja]
    * @description
-   *   [en]Returns whether the OS is iOS. By default does not ignore selected platform.[/en]
+   *   [en]Returns whether the OS is iOS. By default will return manually selected platform if it is set.[/en]
    *   [ja]iOS上で実行されているかどうかを返します。[/ja]
    * @return {Boolean}
    */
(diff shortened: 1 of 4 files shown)


core/src/ons/platform.js
```diff
@@ -138,14 +138,17 @@ class Platform {
...
   /**
    * @method isIOS
-   * @signature isIOS()
+   * @signature isIOS([ignoreSelectedPlatform])
+   * @param {Boolean} ignoreSelectedPlatform
+   *   [en]If true, selected platform is ignored and the actual platform is returned.[/en]
+   *   [ja][/ja]
    * @description
-   *   [en]Returns whether the OS is iOS.[/en]
+   *   [en]Returns whether the OS is iOS. By default does not ignore selected platform.[/en]
```
asialgearoid: I would rephrase this change to not use the word ignore, more like `By default, will return manually selected platform if it is set`

core/src/ons/platform.js
```diff
@@ -138,14 +138,17 @@ class Platform {
...
    *   [ja]iOS上で実行されているかどうかを返します。[/ja]
    * @return {Boolean}
    */
-  isIOS() {
-    if (this._getSelectedPlatform()) {
+  isIOS(ignoreSelectedPlatform = false) {
```
asialgearoid: The `= false` default isn't necessary here, since `undefined` is falsy anyway

core/src/ons/platform.js
```diff
@@ -138,14 +138,17 @@ class Platform {
   //----------------
   /**
    * @method isIOS
-   * @signature isIOS()
+   * @signature isIOS([ignoreSelectedPlatform])
+   * @param {Boolean} ignoreSelectedPlatform
```
asialgearoid: This is a better implementation than I was originally thinking of, nice one. I think this parameter could be renamed though. How about `forceActualPlatform`?

Issue #2479, opened 2018-07-13
Vue: Hidable components should show `visible` prop in docs 


asialgearoid:
Many of our Vue components are "hidable", meaning that the prop `visible` can be set to `false` to hide it. However, the docs for these components do not mention `visible` in their props section. One such example is [v-ons-tabbar](https://onsen.io/v2/api/vue/v-ons-tabbar.html#props).

From what I can see, most of the Vue documentation is taken from the core. I am not certain if our current set up allows us to add Vue-specific props. If not, we need to change that so the docs are more usable.

PR #2480, opened 2018-07-13
feat(ons-toolbar): Add methods to show and hide toolbar

asialgearoid:
This adds similar methods as already exist on `ons-tabbar`. This also works on all the bindings.

While trying to update the docs for Vue, I realised we don't have a way of adding Vue-specific items to the docs. I have created an issue at https://github.com/OnsenUI/OnsenUI/issues/2479.

### DIFF
--- a/bindings/react/src/components/Toolbar.jsx
+++ b/bindings/react/src/components/Toolbar.jsx
@@ -1,4 +1,5 @@
 import SimpleWrapper from './SimpleWrapper.jsx';
+import ReactDOM from 'react-dom';
 
 import PropTypes from 'prop-types';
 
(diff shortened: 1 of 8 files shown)


bindings/react/src/components/Toolbar.jsx
```diff
@@ -33,6 +34,20 @@ class Toolbar extends SimpleWrapper {
...
+
+  componentDidMount() {
+    super.componentDidMount();
+
+    if (this.props.visible !== undefined) {
+      ReactDOM.findDOMNode(this).setVisibility(this.props.visible);
+    }
+  }
+
+  componentWillReceiveProps(nextProps) {
```
emccorson: I just found out `componentWillReceiveProps` is going to get deprecated and renamed to `UNSAFE_componentWillReceiveProps` in React 17. It's fine for React 16 though.

We use this function all over the place so I've created an [issue](https://github.com/OnsenUI/OnsenUI/issues/2482) so we can deal with it all at once.

PR #2506, opened 2018-08-06
Add missing modifier property to Ripple

paulsowden:


emccorson:
Thank you for your contribution, but it looks like this has been already done recently so I'm going to close this PR.

PR #2537, opened 2018-09-12
Prompt dialog should also be removed in Enter events.

laelbe:


laelbe:
https://codepen.io/Lael/pen/ZMovdp
Click prompt button. 
And watch DOM using developer tools.

4 possible events.
1)Click button
2)Press ESC(or back button)
3)Press Enter
4)Click outside of dialog

Issue #2540, opened 2018-09-20
$ons.platform.isIPhoneX() returned false if iPhone XR and iPhone XS Max


tomopy03:
__Environment__

Onsen UI Version:

- 2.10.4

Framework:

- Vue 2.5.17

Framework binding:

- vue-onsenui 2.6.1

Additional libraries:

- 

Platform:

- iPhone XR, Cordova, iOS Simulator
- iPhone XS Max, Cordova, iOS Simulator

__Encountered problem__

$ons.platform.isIPhoneX() returned false.
If iPhone XS, returned true.

__How to reproduce__

masahirotanaka:
Update. We're working on this, so should be ready soon.

asial-matagawa:
@tomopy03 
This problem has been fixed by #2545 and we will the new version of Onsen UI soon.
Thank you for reporting!

PR #2545, opened 2018-09-25
Added Extra sizes for iphone XR #2540

EddieGriffiths:
This fixes the issue where the isIPhoneX() returns false for IPhone XR

### DIFF
--- a/core/src/ons/platform.js
+++ b/core/src/ons/platform.js
@@ -109,9 +109,12 @@ class Platform {
   isIPhoneX() {
     // iPhone 8 and iPhone X have a same user agent. We cannot avoid using window.screen.
     // This works well both in iOS Safari and (UI|WK)WebView of iPhone X.
+    //Added extra sizes for IPhone XR
     return this.isIPhone() &&
-      (window.screen.width === 375 && window.screen.height === 812
-    || window.screen.width === 812 && window.screen.height === 375);
+      (window.screen.width === 375 && window.screen.height === 812 ||
+       window.screen.width === 414 && window.screen.height === 896 ||
+       window.screen.width === 896 && window.screen.height === 414 ||
+       window.screen.width === 812 && window.screen.height === 375);
   }
 
   /**


masahirotanaka:
Thanks! Let us double-check using the actual device.

asial-matagawa:
@EddieGriffiths (Cc: @masahirotanaka )
Thank you!
I will merge your PR and immediately release it in `onsenui@2.10.5` or `onsenui@2.11.0`.

I think this fix is enough for the new 3 iPhone devices XS, XS Max, and XR because:
- The new screen size added in this time is 414 x 896 pt only.
    - X & XS have the same screen size: 375 x 812 pt.
    - XS Max & XR have the same screen size: 414 x 896 pt.
- We still cannot use `cordova-plugin-device` as a way to distinguish X, XS, XS Max, and XR.
    - `window.device.model` will return `iPhone11,2`, `iPhone11,4`(Global)/`iPhone11,6`(China) and `iPhone11,8` on XS, XS Max, and XR respectively ([source](https://ipsw.me/12.0)), but `window.device.model` always returns `x86_64` on their simulators ([source](https://stackoverflow.com/questions/46704773/how-to-detect-iphone-x-ionic-cordova-app/46705305#comment80516039_46705305)).
    - We can choose to use `window.device.model` in the implementation of `ons.platform.isIPhoneX` (only when `cordova-plugin-device` is added) but it is very confusing if the behavior is different between simulators and real devices.
- We also cannot use `navigator.userAgent`.
    - My iPhone XS Max simluator returns `Mozilla/5.0 (iPhone; CPU iPhone OS 12_0 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Mobile/16A366` as the value of `navigator.userAgent`.
    - `16A366` is the model number of iOS 12.0.
    - We still cannot detect the device models by `navigator.userAgent`.
- Conclusion:
    - We still have to depend on `window.screen` to detect the device models as long as we want a simple and comprehensive solution.

PR #2564, opened 2018-10-18
docs(react/Segment): correct name of Segment onChange => onPostChange

aiba:


emccorson:
Thanks for the PR! Sorry for the delay - it's merged now.

Issue #2449, opened 2018-06-20
CSSnext is deprecated.


sbaechler:
Cssnext is no longer being maintained. You might want to look for another solution for the future.

The author recommends [postcss-preset-env](https://preset-env.cssdb.org/)

https://moox.io/blog/deprecating-cssnext/

asialgearoid:
Thanks for the heads up, I hadn't realised that. I guess there is no immediate rush to change, but we should get on this in the next few weeks.

PR #2602, opened 2018-12-19
CSSnext is deprecated

pheadeaux:
Hi, nice framework!

When setting up the `css-components`-folder I am faced with a lot of errors `node-gyp` and `fs-events` running on nodeJs LTS (v10.14.2). But basically the replacement with postcss-preset-env would be easy.

Please confirm.

Related issue: #2449

### DIFF
--- a/css-components/gulpfile.js
+++ b/css-components/gulpfile.js
@@ -8,7 +8,7 @@ const $ = require('gulp-load-plugins')();
 const eco = require('eco');
 const fs = require('fs');
 const ancss = require('@onsenui/ancss');
-const cssnext = require('postcss-cssnext');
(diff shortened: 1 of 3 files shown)


pheadeaux:
I see the actuall problem will be the `@apply` rule, which is not supported by postcss-preset-env.
I don't see any other way than switching to e.g. sass files still using  css-vars...

shrpne:
For the first time, [postcss-apply](https://github.com/pascalduez/postcss-apply) can be used as a standalone postcss plugin without whole `cssnext` pack.
In the future I expect `@apply` rules will be rewriten, as it is out of spec now.

emccorson:
Closing as we are no longer using PostCSS to transform the CSS components (they are written in plain CSS now).

PR #2620, opened 2019-01-21
Add module field in package.json

porsager:
This will allow usage through eg. unpkg.com using `<script src="https://unpkg.com/onsenui?module" type="module"></script>`

It would probably also allow tree shaking to work when bundling with rollup et al.

Be sure to test this with various bundlers before merging ;)
I found [the commit](https://github.com/OnsenUI/OnsenUI/commit/eb909909c49f99c2fc6bb5c573a850527e589364) that removed this saying "temporarily", so perhaps it's time to include again?

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,6 +10,7 @@
     "test:skip-build": "gulp test --disable-warnings --skip-build"
   },
   "main": "js/onsenui.js",
+  "module": "esm/index.js",
   "typings": "js/onsenui.d.ts",
   "files": [
     "bower.json",


asial-matagawa:
Thank you for your PR @porsager!

@emccorson 
Yes, I remember.
There are two main reasons why I removed `core-src/setup.js`:

### Reason 1: `core-src/setup.js` was not pre-transpiled (= written in ES5)

In May 2017 I removed `"module": "core-src/setup.js"` because **`module` files must be pre-transpiled (= written in ES5)** but `core-src/setup.js` was not.
As far as I remember, `core-src/setup.js` caused some critical errors due to ES2015+ syntax and features because **webpack 2+ and Rollup uses `module` files as default but it only simply concatenated the contents of `module` files without any transpilation.**

But now we have `esm/index.js` (not exist in May 2017) which is pre-transpiled, so we can use it as `module` now since `"module": "esm/index.js"` will not cause any problem.

### Reason 2: ES Modules caused critical performance degradation in Monaca CLI

Here is my private Slack log on May 22, 2017:

```
>naoki.m [6:17 PM] 
>@fran @masahiro
>It seems that ES Modules makes build process heavier.
>
>Required time for `monaca preview`:
>- 19.44 s (onsenui 2.2.6 + react-onsenui 1.2.0)
>- 74.00 s (onsenui 2.3.0 + react-onsenui 1.3.0)
>
>Maybe we should force `monaca-lib` to use UMD bundles as default
```

I think we should make sure that ES Modules causes no *critical* performance degradation before merging this PR.
I think 100-150 % performance degradation can be acceptable.

Related: #2001, #2008

PR #2630, opened 2019-02-19
Added missing ">" in Select component, react documentation

Ans0n-Ti0:


emccorson:
Thanks for the PR!

I've merged it now. The changes will show up on onsen.io the next time we do a release of Onsen UI.

Issue #2482, opened 2018-07-13
Remove UNSAFE functions in React


emccorson:
There are some functions in the React bindings which have been deprecated and prefixed with `UNSAFE_`. As of React 17, calling these functions without the `UNSAFE_` prefix will not work.

One example is `componentWillRecieveProps` which should now be called as `UNSAFE_componentWillReceiveProps`.

A quick fix would be to run this https://github.com/reactjs/react-codemod#rename-unsafe-lifecycles which will rename all unsafe functions.

The proper solution is to replace usage of these functions with whatever React recommends instead.

Currently, Onsen UI uses React 16 so this isn't a problem, but it will break if we upgrade to React 17. We should probably remove these functions anyway since React recommends not to use them.

asialgearoid:
This more generally seems to be related to [React's move to asynchronous rendering](https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html), which could require big changes within our bindings.

For now though, I agree we should just use the renamed versions. I can't find any release date for React 17, so I presume it is not imminent.

For maximum compatibility, I would like to keep both functions (both `componentWillRecieveProps` and `UNSAFE_componentWillRecieveProps`). I think we should be able to do this easily inside the components' constructors by setting `this.UNSAFE_componentWillRecieveProps = componentWillRecieveProps`.

emccorson:
bump

PR #2694, opened 2019-09-16
Renamed deprecated react lifecyles to remove the warnings in React 16.9

rj-david:
React 16.9 started to display long warnings in the console regarding deprecated lifecyles. These warnings make development very difficult.

This PR only removes the warnings. It does not remove the deprecated lifecyles. This is a stop-gap solution just to make development easier when using React 16.9. There is still a need to properly update the logic to use the correct lifecyles

### DIFF
--- a/bindings/react/src/components/BackButton.jsx
+++ b/bindings/react/src/components/BackButton.jsx
@@ -38,7 +38,7 @@ class BackButton extends SimpleWrapper {
     this._updateOnClick(this.props);
   }
 
-  componentWillReceiveProps(props) {
+  UNSAFE_componentWillReceiveProps(props) {
     this._updateOnClick(props);
   }
 }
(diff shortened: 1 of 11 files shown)


emccorson:
Thanks for the PR :+1: 

This looks like a good temporary solution, so I will merge it.

We just need to make sure we don't forget to add a permanent solution later; #2482 captures the problem.

amcclain:
Much appreciated, as those warnings are effectively blocking us from updating to the latest react (due to our use of Onsen and several other deps that are in the same boat).

Could you provide any guidance as to when you might cut a versioned release we could use? Thanks!

emccorson:
@amcclain There's not a patch release planned for this month. The earliest will probably be the start of next month but it depends what other fixes get added. You can build your own version of OnsenUI from master following the instructions in CONTRIBUTING.md.

PR #2699, opened 2019-10-06
Update ListItem.jsx

gittysam:
Hello !

Property 'expanded' is not initialized upon component mounting.
This solves a glitch when an expandable ListItem is first tapped : all ListItem expandable siblings get instantly opened and close.

### DIFF
--- a/bindings/react/src/components/ListItem.jsx
+++ b/bindings/react/src/components/ListItem.jsx
@@ -27,6 +27,7 @@ class ListItem extends SimpleWrapper {
   componentDidMount() {
     super.componentDidMount();
     this.node = ReactDOM.findDOMNode(this);
+    this.node.expanded = this.props.expanded === true
   }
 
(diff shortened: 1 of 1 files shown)


emccorson:
Thanks for the PR!

To make sure the list item appears to be expanded, you also need to add the `expanded` class to the node.

Other than that, it all looks good. I've added the extra class bit already so I'll go ahead and merge this now.

gittysam:
Thank you for your answer Eric.

Though it's very humble, this was my first ever contribution ever to open
source !

Also thank you for the great work with OnsenUI.
I'm using it for developing a free web app with tools and materials to help
paediatrics practice, and it is already used by 1000+ physicians in France
each month !

Best regards,

Samuel.


Le ven. 28 févr. 2020 à 05:27, Eric Corson <notifications@github.com> a
écrit :

PR #2712, opened 2019-12-03
chore(css-components): Replace deprecated postcss-cssnext

emccorson:
postcss-apply was also added since postcss-preset-env does not support
the @apply rule:
https://github.com/csstools/postcss-preset-env/issues/50


emccorson:
Closing at PostCSS is long gone.

PR #2720, opened 2020-01-20
Fallback to attr icon when ons-tab updates button content

StaymanHou:
## Expected behavior

When updating the tab attribute, e.g. `badge`, while the tab is active, the icon should fallback to the `icon` attribute if the `active-icon` attribute is not specified.

## Current behavior

In `setActive` method, it behaves as expected by implicitly not updating the icon if the `active-icon` attributed is not specified. However, in `_updateButtonContent` it will set icon to null if the `active-icon` is not specified.

## Temporary solution

I can currently workaround the issue by explicitly set the `active-icon` attribute to the same value as the `icon` attribute. But the proposed the change would make the behavior consistent.


emccorson:
Nice catch! Thanks for the PR!

PR #2721, opened 2020-01-21
add max-height attribute to ons-pull-hook

StaymanHou:
## Expected behavior

When `threshold-height` attribute is a negative value, the pull hook should prevent itself from being pulled too far. And developer should be able to specify the max distance (`max-height` attribute) a user can pull the component

## Current behavior

When `threshold-height` attribute is a negative value, a user can pull the pull-hook component to an indefinite distance which may make the layout look weird.

## Proposed solution

The proposed the change adds an `max-height` attribute to the `ons-pull-hook` component, by which a developer can specify how far a user can pull it when `threshold-height` is disabled.

### DIFF
--- a/bindings/angular1/views/pullHook.js
+++ b/bindings/angular1/views/pullHook.js
@@ -61,7 +61,7 @@ limitations under the License.
 
     MicroEvent.mixin(PullHookView);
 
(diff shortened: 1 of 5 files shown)


StaymanHou:
Hi, just wanted to follow up on this PR. Would the change be reviewed and merged? Thanks

PR #2804, opened 2020-10-09
fix(ons.platform): Fix isIOS returning false on iPad with iOS 14

StaymanHou:


emccorson:
Thanks for the PR! I've merged this now.

PR #2811, opened 2020-11-08
+ Added extra sizes for iPhone 11/12 variants

stonjohnsson:
This fixes the issue where the isIPhoneX() returns false for iPhone 11 & 12 variants. For blank areas on the screen edges - https://onsen.io/v2/guide/iphonex.html

### DIFF
--- a/core/src/ons/platform.js
+++ b/core/src/ons/platform.js
@@ -102,8 +102,8 @@ class Platform {
    * @method isIPhoneX
    * @signature isIPhoneX()
    * @description
-   *   [en]Returns whether the device is iPhone X, XS, XS Max, XR, 11, 12 Mini, 12, 12 Pro or 12 Pro Max.[/en]
-   *   [ja]iPhone X や XS、XS Max、XR、11、12 Mini、12、12 Pro、または12 ProMaxのいずれであるかを返します。[/ja]
+   *   [en]Returns whether the device is iPhone X, XS, XS Max, XR, 11, 11 Pro, 11 Pro Max, 12 Mini, 12, 12 Pro or 12 Pro Max.[/en]
+   *   [ja]iPhone X や XS、XS Max、XR、11、11 Pro、11 Pro Max、12 Mini、12、12 Pro、または12 Pro Max上で実行されているかどうかを返します。[/ja]
    * @return {Boolean}
    */
   isIPhoneX() {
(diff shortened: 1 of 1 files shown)


stonjohnsson:
Nvm: iPhone 11 has the same viewport pixels as X/XS. This fixes only the iPhone 12-variants :-)

mrnebbi:
I was experiencing issues with the new iPhone 12 not positioning toolbars correctly, I suspected it was a safe-area-inset issue, but it was caused by OnsenUI not recognising iPhone 12 as a notched device. Applied this patch and all is good now, thanks @stonjohnsson!

Also, might it be worth deprecating isiPhoneX and and renaming the method to isiPhoneNotched or similar?

core/src/ons/platform.js
```diff
@@ -115,7 +115,13 @@ class Platform {
       (window.screen.width === 375 && window.screen.height === 812 || // X, XS portrait
        window.screen.width === 812 && window.screen.height === 375 || // X, XS landscape
        window.screen.width === 414 && window.screen.height === 896 || // XS Max, XR portrait
-       window.screen.width === 896 && window.screen.height === 414); // XS Max, XR landscape
+       window.screen.width === 896 && window.screen.height === 414 || // XS Max, XR landscape
+       window.screen.width === 360 && window.screen.height === 780 || // 12 Mini portrait
```
emccorson, line 119: Wrong values here but since we already check for the right values above, these can be deleted safely.

emccorson:
@stonjohnsson Thanks for the PR!

I've tested on the 12, 12 Pro, 12 Pro Max, and 12 Mini simulators (no access to a physical device at the moment). It all seems to be working as expected, except for with the 12 Mini.

On the 12 Mini simulator, the width and height returned are `375` and `812`, not `360` and `780`. Since the actual returned values match what we're already checking for the iPhone X, `isIPhoneX` does return true on the 12 Mini simulator.

Before I merge this, can you remove the 12 Mini check with the wrong values and update the comments to include 12 Mini in the iPhone X check as well?


@mrnebbi That is a good point about renaming the function. I will think about what is best to name this and change the name after this PR is merged.

emccorson:
Also I just noticed that the Japanese text has changed significantly. I'm not a native Japanese speaker, so can you please confirm if this is native Japanese? If not (such as if it's from Google translate or similar), I would rather keep the original Japanese text, just adding the new iPhone models to the list.
@stonjohnsson

emccorson:
I've made the changes I suggested and added them to the PR and merged. Thank you for your contribution :tada:

stonjohnsson:
Super! I didn't have time to respond & make the adjustments. Thank you for doing this.

PR #2850, opened 2021-02-04
Use Vue 3 for examples app

emccorson:
Updates the examples app to use Vue 3. Bumps Vue dependencies and fixes code to use Vue 3 syntax.


emccorson:
Not going to merge. Replaced with #2850 which is the same code but with tidier commits and a better branch name (because this PR doesn't actually fix issue-2832 even though that is the branch name).

PR #2886, opened 2021-04-01
perf(core): call DOMContentLoaded and deviceready just once

Sysix:
we don't need to listen for a second call, should never happend.


emccorson:
Thanks for the PR! I've checked it and it all looks good. All I've added is a commit to update the CHANGELOG and CONTRIBUTORS files. I'll merge this now.

PR #2932, opened 2021-08-10
Publish from root directory

### LINKED ISSUE #2926: Consider dropping OnsenUI-dist directory
When distributing Onsen UI, we run `npm run dist` which creates a new directory `OnsenUI-dist` containing just the files that should be distributed. `npm publish` is run from the `OnsenUI-dist` directory, _not_ from the root directory.

This can be a bit complicated and we could simplify things by publishing directly from the root directory (like a lot of other npm packages do).

This would require setting the `files` key in `package.json` correctly to make sure we only publish the files we want to publish.

Unfortunately, this would probably mean a change to the distributed directory structure of Onsen UI. I'm thinking that we should output the build to a new directory `dist`, which would mean that the location of Onsen UI would be `dist/js/onsenui.js` instead of `js/onsenui.js` as it currently is. This would be a breaking change which would mainly impact users installing Onsen UI using a `<script>` tag in their HTML.

Consider whether this is a good idea and if so do it.

emccorson:
Builds and publishes from the root directory instead of `build` and `OnsenUI-dist`.

Changes the directory structure:
```
build/js --> js
core/esm --> esm
build/esm --> we no longer build ESM
css-components/src --> css/css-components-src
core/css/src --> css/src
build/css --> css
OnsenUI-dist --> unused. npm publish should now be run from the root directory
```

This PR also merges css-components and the core css to use the same postcss config.

Fixes #2926.


emccorson:
Pausing work on this PR until #2933 is dealt with. In order to publish ESM without a build step, we need to get rid of the Custom Elements polyfill dependency, and this requires fixing the codebase as per #2933 (probably by moving to the Shadow DOM).

PR #2963, opened 2022-03-23
Update AlertDialogButton.jsx

vai12376:


emccorson:
I'm not sure what the intention of this PR is, but as it currently stands the code is not valid JavaScript and it appears to be a test PR, so I will close this.

Issue #2964, opened 2022-03-30
the app crash when push a page that not exists


happyi:
__Environment__

Onsen UI Version:
v2.11.1
- 

Framework:
angularjs
1.4.7

Framework binding:
v1.0.0
- 

Platform:
all

use  ons-navigator  

1 :   push a a page that not exists  
 ```
     appNavigator.push('app/home/log.xml')
```
2:  then push or pop a normal  page  it get's

```
     framework.min.js:297 Uncaught (in promise) popPage is already running.
     framework.min.js:297 Uncaught (in promise) pushPage is already running.

```

it's look like crashed  by the end-user。

suggest：
      set a timeout let appNavigator return Initial State anyway

**Comment by emccorson** [comment]
I guess the fix here is to provide a better error message.
Although angularjs is mentioned in the issue, this is actually a problem with the core.

I have provided an example to reproduce that can be pasted in the playground.

```html
<ons-navigator swipeable id="myNavigator" page="page1.html"></ons-navigator>

<template id="page1.html">
  <ons-page id="page1">
    <ons-toolbar>
      <div class="center">Page 1</div>
    </ons-toolbar>

    <p>This is the first page.</p>

    <ons-button id="push-button">Push page</ons-button>
  </ons-page>
</template>
```

```javascript
document.addEventListener('init', function(event) {
  document.querySelector('#push-button').onclick = function() {
    document.querySelector('#myNavigator').pushPage('page2.html');
  };
});
```

yong-asial:
@emccorson I'd like to attempt solving this issue. I am appreciated if you could point out where (related files) should i look at?

emccorson:
@yong-asial Thanks for choosing this issue. The place to look is onsenui/esm/elements/ons-navigator.js, and specifically the `pushPage` function. That function calls the page loader to load the page, and I think an error is returned there when the page is not found.

You can also paste the example from my previous comment into onsen.io/playground and follow the stack trace in the console to find the source of the error.

I would adding that example to the `onsenui/examples` directory during development, then run `npm run serve` so you can see the problem locally. When you edit `ons-navigator.js` it should reload the examples.

Also needed would be to add a unit test to `ons-navigator.spec.js`.

Good luck!

yong-asial:
@emccorson Thank you. 

[...]

PR #3014, opened 2022-09-22
fix(ons-navigator): reject proper message (404) when push none existing page

yong-asial:
- reject Error instead of throws error.
- add test to handle none existing page.
- pageLoader: return 404 if the page is not existed.
- add button to push none existing page in the example.

Related issue: #2964


emccorson (approved):
This is nice! I will merge this now.

PR #3028, opened 2022-09-29
feat(ons-list-item): Add keep-tap-background-color property

### LINKED ISSUE #2626: React | List issue: Dynamically set backgroundColor is reset
__Environment__

  
```
[Core]
  onsenui latest
  
[Framework]
  react 18.2.0

[Framework binding]
 react-onsenui latest

[Platform]
  Desktop - Ubuntu 64-bit

[Browser]
  Desktop - Firefox 65.0 on Ubuntu 64-bit
```

__Encountered problem__
  

__How to reproduce__
  

  
  [__Demo link__](https://onsen.io/playground/?issue)

  

- __HTML__

```html
<div id="app"></div>
```

- __JS__

```jsx
class MyPage extends React.Component {
  renderToolbar() {
    return (
      <Ons.Toolbar>
        <div className='center'>List</div>
      </Ons.Toolbar>
    );
  }

  renderRow(row, index) {
    const x = 40 + Math.round(5 * (Math.random() - 0.5)),
          y = 40 + Math.round(5 * (Math.random() - 0.5));

    const names = ['Max', 'Chloe', 'Bella', 'Oliver', 'Tiger', 'Lucy', 'Shadow', 'Angel'];
    const name = names[Math.floor(names.length * Math.random())];
    
    const style = {}
    if(this.state && index === this.state.clicked)
      style.backgroundColor = 'green';

    return (
      <Ons.ListItem
        key={index}
        onClick={() => {this.setState({clicked: index})}}
        tappable={true}
        style={style}>
        <div className='left'>
          <img src={`http://placekitten.com/g/${x}/${y}`} className='list-item__thumbnail' />
        </div>
        <div className='center'>
          {name}
        </div>
      </Ons.ListItem>
    );
  }

  render() {
    return (
      <Ons.Page renderToolbar={this.renderToolbar.bind(this)}>
        <Ons.List
          dataSource={[1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12]}
          renderRow={this.renderRow.bind(this)}
          renderHeader={() => <Ons.ListHeader>Cute cats</Ons.ListHeader>}
        />
      </Ons.Page>
    );
  }
}

ons.ready(function() {
  ReactDOM.createRoot(document.getElementById('app')).render(<MyPage />);
});
```

When a user clicks on a list item, I try to dynamically change the background color of that item. While the above code works to set the background color, it gets unset by an event within Onsen `mouseout --> _onRelease`, which reverts the background color to the "original" background color when the mouse moves off the list item. From the web console, events `mouseup` and `mouseout` both call `_onRelease`. If I'm not mistaken, only `mouseup` needs to be doing so to ensure dynamically set background colors are unaffected

My code above is a modified version of the playground code. If you click on an item, it'll set the background color and will only reset it once your mouse moves off the item.

Is there any way for me to accomplish what I need? I can't use CSS classes because the color is dynamically determined. I think this is an unconsidered edge case since the user could theoretically leave their mouse on the list item for an indefinite amount of time while the background color changes dynamically.

as-hirose:
Added `keep-tap-background-color` property to prevent from clearing the background color on `"touchmove"`, `"touchcancel"`, `"touchend"`, `"touchleave"`, `"mouseup"`, and `"mouseout"`. For this to work, the attribute "tappable" needs to be set.

Fixes #2626.

### DIFF
--- a/onsenui/esm/elements/ons-list-item.js
+++ b/onsenui/esm/elements/ons-list-item.js
@@ -159,6 +159,22 @@ export default class ListItemElement extends BaseElement {
    */
 
   /**
+   * @attribute keep-tap-background-color
+   * @type {Boolean}
+   * @description
+   *  [en] Prevent from clearing the background color on `"touchmove"`, `"touchcancel"`, `"touchend"`, `"touchleave"`, `"mouseup"`, and `"mouseout"`. For this to work, the attribute "tappable" needs to be set.[/en]
+   *  [ja] この属性があると、`"touchmove"`, `"touchcancel"`, `"touchend"`, `"touchleave"`, `"mouseup"`, and `"mouseout"` イベント時に背景色がクリアされないようになります。[/ja]
+   */
+
+  /**
+   * @property keepTapBackgroundColor
+   * @type {Boolean}
+   * @description
+   *   [en] Prevent from clearing the background color on `"touchmove"`, `"touchcancel"`, `"touchend"`, `"touchleave"`, `"mouseup"`, and `"mouseout"`. For this to work, the attribute "tappable" needs to be set.[/en]
+   *   [ja] この属性があると、`"touchmove"`, `"touchcancel"`, `"touchend"`, `"touchleave"`, `"mouseup"`, and `"mouseout"` イベント時に背景色がクリアされないようになります。[/ja]
+   */
+
+  /**
    * @attribute expandable
    * @type {Boolean}
    * @description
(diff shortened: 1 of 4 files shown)


emccorson (changes requested):
Sorry, I forgot about the changes that will be needed to the React component now.
Can you please also add the prop `keepTapBackgroundColor` to react-onsenui/src/components/ListItem.jsx?
All you need to do is add a new prop to the `propTypes` object. This is just for documentation; the prop will automatically be passed to the core as an attribute.

Can you please also add the changes to onsenui/CHANGELOG.md and react-onsenui/CHANGELOG.md? For any user-facing changes we need to add it to the changelog.

Apart from that, everything looks good! Just make the changes and squash them.

as-hirose:
@emccorson Thank you very much. I added `keepTapBackgroundColor` and squashed the commits. I will update CHANGELOG when I release this.

Issue #3076, opened 2024-02-15
Wrong page element returned, when inserting several pages


magynhard:
__Environment__

Onsen UI Version:

- 2.12.8

Framework:

- JavaScript

Framework binding:

- none

Additional libraries:

- none

Platform:

- Windows 10 Chome

__Encountered problem__

When inserting several Pages at the same time, the resolved Promise does not return the correct page elements.

__How to reproduce__

```js
['my-page-1.html', 'my-page-2.html', 'my-page-3.html'].forEach((page) => {
   document.querySelector('ons-navigator').insertPage(0, page).then((el) => {
     console.log(el); // always returns my-page-3
   });
})
```

The problem is, that the insertPage method return pages[index] instead of the actual page element. When pages are added at the "same" time, like a loop, then the last element "wins", as there is some delay between the promise resolving.

PR with fix will follow within a few minutes.

magynhard:
As the project seems not to be maintained anymore, i added this patch and others on my fork:
https://github.com/magynhard/OnsenUI

PR #3077, opened 2024-02-15
Ensure correct page return, when inserting several pages at once

magynhard:
Fixes issue #3076 by resolving the actual page element instead of the index which is wrong in the described use case.


magynhard:
As the project seems not to be maintained anymore, i added this patch and others on my fork:
https://github.com/magynhard/OnsenUI

